PR #117, opened 2023-05-19
Algo: Add CCTK_HOST CCTK_DEVICE to function brent and newton_raphson_nd

### LINKED ISSUE #108: Add CCTK_HOST and CCTK_DEVICE for functions in Algo
Should we add CCTK_HOST and CCTK_DEVICE in front of those functions in Algo?
In AsterX/Con2PrimFactory, we called function `brent()` which requires us to add CCTK_HOST and CCTK_DEVICE to make it work with GPU.

lwJi:
Closes #108.

### DIFF
--- a/Algo/src/roots.cxx
+++ b/Algo/src/roots.cxx
@@ -16,10 +16,11 @@ namespace Algo {
 using Arith::mat, Arith::vec;
 
 namespace {
-template <typename T> vec<T, 2> gn(vec<T, 2> x) {
+template <typename T> CCTK_HOST CCTK_DEVICE vec<T, 2> gn(vec<T, 2> x) {
   return vec<T, 2>{x(0) * x(0) - 2, x(0) * x(1) - 2};
 }
-template <typename T> std::pair<vec<T, 2>, mat<T, 2> > gnd(vec<T, 2> x) {
+template <typename T> CCTK_HOST CCTK_DEVICE
+std::pair<vec<T, 2>, mat<T, 2> > gnd(vec<T, 2> x) {
   return {vec<T, 2>{x(0) * x(0) - 2, x(0) * x(1) - 2},
           mat<T, 2>{2 * x(0), x(1), 0, x(0)}};
 }
@@ -28,9 +29,11 @@ template <typename T> std::pair<vec<T, 2>, mat<T, 2> > gnd(vec<T, 2> x) {
 extern "C" void Test_roots(CCTK_ARGUMENTS) {
   DECLARE_CCTK_ARGUMENTS;
 
-  auto fn = [](auto x) { return x * x - 2; };
-  auto fnd = [](auto x) { return std::make_tuple(x * x - 2, 2 * x); };
-  auto fnd2 = [](auto x) { return std::make_tuple(x * x - 2, 2 * x, 2); };
+  auto fn = []CCTK_HOST CCTK_DEVICE (double x) { return x * x - 2; };
+  auto fnd = []CCTK_HOST CCTK_DEVICE (double x) {
+    return std::make_tuple(x * x - 2, 2 * x); };
+  auto fnd2 = []CCTK_HOST CCTK_DEVICE (double x) {
+    return std::make_tuple(x * x - 2, 2 * x, 2); };
 
   {
     const int minbits = std::numeric_limits<double>::digits - 4;
@@ -57,19 +60,6 @@ extern "C" void Test_roots(CCTK_ARGUMENTS) {
     assert(hi >= lo && hi - lo <= std::scalbn(2.0, -minbits));
     assert(fn(lo) <= 0 && fn(hi) >= 0);
     CCTK_VINFO("Test_bracket_and_solve_root succeeded in %d iterations", iters);
-  }
-
-  {
-    const int minbits = std::numeric_limits<double>::digits - 4;
-    const int maxiters = 100;
-    int iters;
-    auto [lo, hi] = brent(fn, 1.0, 2.0, minbits, maxiters, iters);
-    // CCTK_VINFO("maxiters=%d iters=%d", maxiters, iters);
-    // CCTK_VINFO("lo=%.17g hi=%.17g", double(lo), double(hi));
-    assert(iters < maxiters);
-    assert(hi >= lo && hi - lo <= std::scalbn(2.0, -minbits));
-    assert(fn(lo) <= 0 && fn(hi) >= 0);
-    CCTK_VINFO("Test_brent succeeded in %d iterations", iters);
   }
 
   {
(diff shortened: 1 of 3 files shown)


eschnett:
@lwJi Did you try using these functions in a GPU kernel? Can you set up a test case that would call them? There does not need to be a run-time test (because we currently cannot run tests on GPUs), but compile time tests would be good.

lwJi:
I have tried `brent()` in a GPU kernel and it works perfectly. I haven't test `newton_raphson` yet. And the reason I haven't add CCTK_DEVICE for the other functions is they are still using `boost::`.

Aha. I misunderstood what you said before. I will try to add a `Test_roots`.

lwJi:
@eschnett I think the tests are already exist in `root.cxx`
https://github.com/eschnett/CarpetX/blob/b68c3ae7fc64323bf0a2b007a3db4624b5d30f50/Algo/src/roots.cxx#L66

eschnett:
These functions are built for the CPU only, not for a GPU.

lwJi:
Now I get it (finally). Let me try again ;)

lwJi:
@eschnett I think I have finished adding GPU tests (for `brent` and `newton_raphson_nd` only for the moment). Please let me know if I missed anything else

Algo/test/test_roots.par
```diff
@@ -1,2 +1,8 @@
-ActiveThorns = "Algo"
+ActiveThorns = "
+    CarpetX
+    IOUtil
+    Algo
```
eschnett, line 4: Can you sort the thorn names alphabetically?

Algo/src/roots.cxx
```diff
@@ -28,9 +29,11 @@ template <typename T> std::pair<vec<T, 2>, mat<T, 2> > gnd(vec<T, 2> x) {
 extern "C" void Test_roots(CCTK_ARGUMENTS) {
   DECLARE_CCTK_ARGUMENTS;
 
-  auto fn = [](auto x) { return x * x - 2; };
-  auto fnd = [](auto x) { return std::make_tuple(x * x - 2, 2 * x); };
-  auto fnd2 = [](auto x) { return std::make_tuple(x * x - 2, 2 * x, 2); };
+  auto fn = []CCTK_HOST CCTK_DEVICE (double x) { return x * x - 2; };
```
eschnett, line 32: Can you use `auto` (or `CCTK_REAL`) instead of `double`?

Algo/src/roots.cxx
```diff
@@ -115,21 +105,45 @@ extern "C" void Test_roots(CCTK_ARGUMENTS) {
...
-                               vec<CCTK_REAL, 2>{10.0, 10.0}, minbits, maxiters,
-                               iters, failed);
-    // CCTK_VINFO("maxiters=%d iters=%d", maxiters, iters);
-    // CCTK_VINFO("lo=%.17g hi=%.17g", double(lo), double(hi));
-    assert(iters < maxiters);
-    assert(sumabs(gn(x)) < 1.0e-9);
-    CCTK_VINFO("Test_newton_raphson_nd succeeded in %d iterations", iters);
+    const amrex::Box box(amrex::IntVect(0), amrex::IntVect(0, 0, 0));
+
+    amrex::ParallelFor(
```
eschnett, line 110: Use `amrex::launch` instead; see https://amrex-codes.github.io/amrex/docs_html/GPU.html#launching-general-kernels .

Algo/src/roots.cxx
```diff
@@ -115,21 +105,45 @@ extern "C" void Test_roots(CCTK_ARGUMENTS) {
...
+      const int maxiters = 100;
+      int iters;
+      auto [lo, hi] = brent(fn, 1.0, 2.0, minbits, maxiters, iters);
+      // CCTK_VINFO("maxiters=%d iters=%d", maxiters, iters);
+      // CCTK_VINFO("lo=%.17g hi=%.17g", double(lo), double(hi));
+      assert(iters < maxiters);
+      assert(hi >= lo && hi - lo <= std::scalbn(2.0, -minbits));
+      assert(fn(lo) <= 0 && fn(hi) >= 0);
+      //CCTK_VINFO("Test_brent succeeded in %d iterations", iters);
+      printf("INFO (Algo): Test_brent succeeded in %d iterations\n", iters);
```
eschnett, line 123: Don't call `printf` from inside the kernel. Instead, check whether the kernel finished successfully outside the kernel, and then call `CCTK_INFO`.

Algo/src/roots.cxx
```diff
@@ -16,10 +16,11 @@ namespace Algo {
 using Arith::mat, Arith::vec;
 
 namespace {
-template <typename T> vec<T, 2> gn(vec<T, 2> x) {
+template <typename T> CCTK_HOST CCTK_DEVICE vec<T, 2> gn(vec<T, 2> x) {
```
eschnett, line 19: If these functions are only called inside a kernel then the `CCTK_HOST` declaration is not necessary.

Algo/src/roots.hxx
```diff
@@ -1,6 +1,7 @@
 #ifndef ALGO_HXX
 #define ALGO_HXX
 
+#include "loop.hxx"
```
eschnett, line 4: Can you write `<loop.hxx>`, and sort the statements alphabetically?

eschnett (changes requested):

Algo/src/roots.cxx
lwJi: should I change all the `double` to `CCTK_REAL` which appears in the file

lwJi:
@eschnett I have change the PR accordingly, please take another look. Thanks

Algo/src/roots.cxx
```diff
@@ -114,22 +104,50 @@ extern "C" void Test_roots(CCTK_ARGUMENTS) {
...
-    auto x = newton_raphson_nd(gnd<CCTK_REAL>, vec<CCTK_REAL, 2>{1.0, 1.0},
-                               vec<CCTK_REAL, 2>{0.0, 0.0},
-                               vec<CCTK_REAL, 2>{10.0, 10.0}, minbits, maxiters,
-                               iters, failed);
-    // CCTK_VINFO("maxiters=%d iters=%d", maxiters, iters);
-    // CCTK_VINFO("lo=%.17g hi=%.17g", double(lo), double(hi));
-    assert(iters < maxiters);
-    assert(sumabs(gn(x)) < 1.0e-9);
-    CCTK_VINFO("Test_newton_raphson_nd succeeded in %d iterations", iters);
+    amrex::launch(box, [=] CCTK_DEVICE(amrex::Box const& tbx)
```
eschnett, line 112: I think that launching a GPU kernel does not wait for the kernel to finish running. Can you check, and add the respective wait statement if necessary?

eschnett (changes requested):

Algo/src/roots.cxx
lwJi: there is an implicit sync in `copyToHost`
```
    T* copyToHost ()
    {
#ifdef AMREX_USE_GPU
        if (d_data)
        {
            Gpu::dtoh_memcpy_async(h_data, d_data, m_size*sizeof(T));
            Gpu::streamSynchronize();
        }
#endif
        return h_data;
    }
```

eschnett (approved):

PR #140, opened 2023-06-06
Algo: declare dependency on Loop

rhaas80:
and thus CarpetX and thus AMReX

### DIFF
--- a/Algo/configuration.ccl
+++ b/Algo/configuration.ccl
@@ -2,6 +2,7 @@
 
 REQUIRES Arith
 REQUIRES Boost
+REQUIRES Loop
 
 PROVIDES Algo
 {


eschnett:
I don't think that `Algo` depends on `Loops`. It doesn't include `loop.hxx`, nor does it contain loops over grid functions.

lwJi:
It does. since `CCTK_HOST` and `CCTK_DEVICE` are defined there

eschnett:
`Loop` is a heavy dependency. I would rather not have it just for some macros.

We can define these macros elsewhere, e.g. in `defs.hxx` of `Arith`. For example, `Arith` already defines `ARITH_DEVICE` etc., and we could extend this to define `CCTK_DEVICE` as well. Or we move these definitions to the flesh. Or we define `ALGO_DEVICE` locally.

rhaas80:
Well I did not do a very careful check where the dependency comes from. The error message produced was:  
\`\`\`  
In file included from /data/rhaas/postdoc/gr/cactus/CarpetX\_SPIN/configs/sim/bindings/include/loop.hxx: ,  
                 from /data/rhaas/postdoc/gr/cactus/CarpetX\_SPIN/arrangements/CarpetX/Algo/src/roots.hxx:4,  
                 from /data/rhaas/postdoc/gr/cactus/CarpetX\_SPIN/configs/sim/build/Algo/roots.cxx:1:  
/data/rhaas/postdoc/gr/cactus/CarpetX\_SPIN/arrangements/CarpetX/Loop/src/loop.hxx:4:10: fatal error: AMReX\_FArrayBox.H: No such file or directory  
    4 | #include \<AMReX\_FArrayBox.H>  
\`\`\`  
which I took to mean that Alog's `root.cxx` includes `loop.hxx` and thus added the `Lopp` dependency. I agree that this would be quite a heavy dependency for Algo, which would otherwise not be dependent on a mesh and implement generic algorithms.

eschnett:
I see. The test cases are calling into AMReX, and thus the thorn needs to depend on the `AMReX` capability. In an ideal world the test cases could be moved to a separate thorn, but I doubt anyone will use Algo without AMReX in the foreseeable future.

Can we remove the dependency on `loop.hxx` in `roots.hxx` instead?

We should also add the missing include statement to `roots.cxx` to define the respective AMReX functions.

lwJi:
Sure. How about define `ALGO_DEVICE` locally then?

eschnett:
I suggest https://github.com/eschnett/CarpetX/pull/141 instead of this pull request. This adds a dependency on AMReX and defines `ALGO_DEVICE` etc. until respective flesh definitions become available.

rhaas80:
sure. I took a look and approved it.

PR #149, opened 2023-06-21
Removed duplicated scheduling

SamuelCupp:
When ODESolvers is on (basically everything right now), several groups are double-scheduled in ADMBase, HydroBase, and TmunuBase. This simply puts in logic that switches their scheduling based on whether this thorn is active. If several thorns like ODESolvers eventually exist, then we might need to introduce a more elegant solution.


SamuelCupp:
@eschnett I have updated the fork in the PR to be up-to-date with the master repo. The tests are still running, but this should be ready to merge.

PR #175, opened 2023-07-13
Loop: reordered loop_box arguments to be more standard

SamuelCupp:
loop_box and loop_box_device both had the function argument 'f' first, while all other functions have this argument last. This changes them to match.

### DIFF
--- a/Loop/src/loop.hxx
+++ b/Loop/src/loop.hxx
@@ -145,10 +145,10 @@ public:
 
   // Loop over a given box
   template <int CI, int CJ, int CK, int VS = 1, typename F>
-  void loop_box(const F &f, const vect<int, dim> &restrict bnd_min,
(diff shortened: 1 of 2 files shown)


eschnett:
@SamuelCupp PR https://github.com/eschnett/CarpetX/pull/174 (just merged) should include your changes. Can you confirm?

SamuelCupp:
It looks like it contains all the changes in the code that is compiled. However, `loop_box_boundary` in `loop.hxx` was not reordered to have the `f` argument last. Similarly, `loop_box_boundary_device` in `loop.hxx` removed the `f` argument from the front but didn't add it to the end of the argument list. These are both currently surrounded by `#if 0` logic, so they aren't compiled.

eschnett:
Applied in https://github.com/eschnett/CarpetX/pull/177.

PR #180, opened 2023-07-24
Derivs: Use GF3D2 as input in calc_derivs

lwJi:


Derivs/src/derivs.cxx
```diff
@@ -10,9 +10,29 @@ using namespace Loop;
 
 template <int CI, int CJ, int CK, typename T>
 CCTK_ATTRIBUTE_NOINLINE void
-calc_derivs(const vec<GF3D5<T>, dim> &dgf, const GridDescBaseDevice &grid,
-            const GF3D5<const T> &gf, const GF3D5layout layout,
-            const vect<T, dim> dx, const int deriv_order) {
+calc_copy(const GF3D5<T> &gf0, const GF3D5layout layout,
```
eschnett, line 13: The output variables have the suffix `0`, the input variables the suffix `1`. That's a bit weird. Can you use `0` for the input variables, and use no suffix for the output variables?

Derivs/src/derivs.hxx
```diff
@@ -332,19 +332,18 @@ inline CCTK_ATTRIBUTE_ALWAYS_INLINE
...
   using namespace Loop;
   // We use explicit index calculations to avoid unnecessary integer
   // multiplications
-  const T *restrict const ptr = &gf(layout, I);
+  const T *restrict const ptr = &gf(I);
   const std::array<std::ptrdiff_t, Loop::dim> offsets{
-      layout.delta(1, 0, 0),
-      layout.delta(0, 1, 0),
-      layout.delta(0, 0, 1),
+      gf.delta(1, 0, 0),
```
eschnett, line 344: I don't think these lines are correct. I don't think that the grid function class `GF3D2` has a `delta` member function.

eschnett (changes requested):
I added comments requesting changes.

Derivs/src/derivs.cxx
lwJi: I modify suffix as suggested.

lwJi:
Hi @eschnett , I have addressed your comments. please take a look again. Thanks

Derivs/src/derivs.hxx
eschnett: nice!

eschnett (approved):

PR #188, opened 2023-07-27
New Loop Defaults

stevenrbrandt:
Change the default layout name
include mask in PointDesc
add decls for CCTK_REALVEC,
CCTK_BOOLVEC, and CCTK_VECSIZE.


eschnett:
Merged as https://github.com/eschnett/CarpetX/pull/202.

PR #190, opened 2023-07-28
Loop: Add new loops for radiative boundary conditions

chcheng3:
This adds the loop_radbnd and loop_radbnd_device functions. These loops are done over the "radiative boundary", which is the outer boundary shrunk inward by the number of ghost points into the interior, so that NewRad can access the points with valid data in the interior.


Loop/src/loop.hxx
```diff
@@ -517,6 +521,110 @@ public:
     } // for rank
   }
 
+  // Alex Dima:
```
eschnett, line 524: Don't put your name there. If your name is there, then others will be afraid of changing the code because they will defer to your authority. To find out who wrote what code one can use `git blame`.

Loop/src/loop.hxx
```diff
@@ -517,6 +521,110 @@ public:
...
+  // Modified from loop_bnd
+  // Loop over all "radiative" boundary points.
```
eschnett, line 526: I think this function is useful not just for radiative boundaries but for all kinds of evolved boundary conditions. Let's use a more general name.

Loop/src/loop.hxx
```diff
@@ -517,6 +521,110 @@ public:
...
+
+                const vect<int, dim> inormal{ni, nj, nk};  //define normal vector
+
+                vect<int, dim> imin, imax;
+                for (int d = 0; d < dim; ++d) { // Set imin, imax in each direction
+                  switch (inormal[d]) {
+                  case -1: // lower boundary
+                    // shift "upwards" from physical boundary by # of ghostpoints
+                    //                                      // xxx|~~~~~~~~~~~~~~|~~~
+                    imin[d] = all_min[d] + nghostzones[d];  // ~~~ ^~~~~~~~~~~~~~ ~~~
```
eschnett, line 589: It seems that `all_min` and `int_min` are always shifted upwards, and `all_max` and `int_max` are always shifted downwards. Instead of using this `for` loop with the `switch` logic, you could just write `imin = all_min + nghostzones` etc., similar to the way `bnd_min` is handled below.

Loop/src/loop.hxx
```diff
@@ -517,6 +521,110 @@ public:
...
+                  // Adjust loop_box index range according to tile min, max
+                  imin[d] = max(tmin[d], imin[d]);
+                  imax[d] = min(tmax[d], imax[d]);
+                }
+
+                // Provide loop_box with bnd_min and bnd_max shifted my # of ghostpoints
+                // so that p.NI can be assigned correct values on the "radiative" boundary.
+                // (NewRad requires knowledge of the normal vectors to its boundary points)
+                loop_box<CI, CJ, CK, VS, N>
+                  (bnd_min+nghostzones, bnd_max-nghostzones, imin, imax, f);
```
eschnett, line 618: Can you run `clang-format` on the source file?

Loop/src/loop_device.hxx
```diff
@@ -88,6 +88,9 @@ public:
           const vect<int, dim> NI =
               vect<int, dim>(I > bnd_max1 - 1) - vect<int, dim>(I < bnd_min1);
           const vect<int, dim> I0 =
+          // If NI<0,   retrun bnd_min1
```
eschnett, line 91: `retrun` -> `return`

Loop/src/loop_device.hxx
```diff
@@ -261,6 +264,111 @@ public:
     } // for rank
   }
 
+  // Alex Dima:
```
eschnett, line 267: Same comments as for the other source file apply here.

eschnett (changes requested):
Thanks for this function! This is a good idea.

Can you add a test case as well?

Loop/src/loop.hxx
chcheng3: Good idea, do you have any suggestions for a more general name?

eschnett: You could ask on the ET channel.

Loop/src/loop.hxx
chcheng3: That is a good point, I  removed it from the comment.

Loop/src/loop.hxx
chcheng3: On the CarpetX gitter chat, Liwei Ji suggested `loop_int_outermost` and `loop_int_bnd`. Since it is not only the outermost single layer of points that gets looped over, but rather the outermost layer of thickness `nghostzones`, I prefer `loop_int_bnd` slightly over `loop_int_outermost`. However, it could be unclear what exactly int_bnd refers to without using a more specific name. What do you think?

Loop/src/loop.hxx
chcheng3: That is a good observation. I updated it to apply the upward shifts to `all_min` and `int_min`, and downward shifts to `all_max` and `int_max` all in one place. However, it seems to me that the switch is still necessary, depending on whether the point lies on the upper/lower boundary or interior. If there is another way to do it without the switch, could you spell it out more explicitly for me?

Loop/src/loop.hxx
chcheng3: Yes, I'll make sure to run it before making commits.

Loop/src/loop_device.hxx
chcheng3: Thanks for spotting the typo!

chcheng3:
I have addressed your comments in the last few commits. In addition to the above changes, I also added a new thorn `TestLoop` with a test parfile that checks whether the loops are done in the correct grid points. Could you please take a look and let me know what you think?

Loop/src/loop.hxx
rhaas80: something like:

```
clang-format-14 -style file -i Loop/src/loop.hxx
```

where the `file` style makes it use the `.clang-format` file that Cactus brings with it.

You may have to check which `clang-format-NN` verison to use since to have it only reformat your changed lines. They are, unfortunately, not stable across versions (and we do not want a formatting war between different versions). 

Cactus prescribes clang-format-9 but CarpetX has been using a newer version (14 seemed to have worked for me in the past).

`git add -p` is also very helpful to pick only reformatting of your own source code (instead of the whole file).

Loop/src/loop_device.hxx
```diff
@@ -261,6 +264,108 @@ public:
...
+    //   all_min = ghost_offset;
+    //   all_max = lsh - offset - ghost_offset;
+    // Indices for the interior
+    //   int_min = nghostzones;
+    //   int_max = lsh - offset - nghostzones;
+    domain_boxes<CI, CJ, CK>(group_nghostzones, all_min, all_max, int_min,
+                             int_max);
+
+    // Shift the indices towards the interior by nghostzones
+    all_min += nghostzones;
```
eschnett, line 304: The variable `nghostzones` contains the default number of ghost zones. Individual grid functions can override this, and the actual number of ghost zones for the loop is passed in `group_nghostzones`. In most cases both with be the same.

I don't think you're handling the case correct when these values are different. One way to "fix" this would be to check whether they are the same, and to abort if they are not.

Loop/src/loop.hxx
```diff
@@ -527,6 +531,105 @@ public:
...
+  // Loop over the outermost "boundary" points in the interior. They correspond
+  // to points that are shifted inwards by = cctk_nghostzones[3] from those that
+  // CarpetX identifies as boundary points. From the perspective of CarpetX (or
+  // AMReX), these do not belong in the outer boundary, but rather the interior.
+  // This excludes ghost faces, but includes ghost edges/corners on non-ghost
+  // faces. Loop over faces first, then edges, then corners. Modified from
+  // loop_bnd.
+  template <int CI, int CJ, int CK, int VS = 1, int N = 1, typename F>
+  inline CCTK_ATTRIBUTE_ALWAYS_INLINE void
+  loop_int_bnd(const vect<int, dim> &group_nghostzones, const F &f) const {
```
eschnett, line 543: I have been thinking about the name a bit, and I now think that `int_bnd` would be confusing. This sounds as if this loop was looping over the interior and the boundary points, which it isn't. Can you change this to `outermost_int` or similar?

eschnett (changes requested):

Loop/src/loop_device.hxx
chcheng3: I just added ```assert(group_nghostzones[d] == nghostzones[d]``` to check for this.

Loop/src/loop.hxx
chcheng3: Hmm, I see where it could be confusing. I have renamed the loops to `outermost_int`.

eschnett (approved):

PR #196, opened 2023-08-05
Derivs: Remove mask when loadu in calc_deriv2

lwJi:


eschnett:
Why do you want to remove the mask?

eschnett:
I think the mask should still be there, but the value of `imax` should be increased by the stencil radius.

lwJi:
> Why do you want to remove the mask?

Some of grid points which are required to construct FD derivatives are masked with `false`, so the values used is always `0.0`. And this case only happened when x-dir is divided into multiple tiles. Here by removing `mask` I was testing if it works when I always keep `mask true`.

lwJi:
> I think the mask should still be there, but the value of `imax` should be increased by the stencil radius.

I can try it out. Could you explain more about `const vbool mask = mask_for_loop_tail<vbool>(p.i, p.imax);`?
when `mask = true`, does it mean `p.i` is inside `(0, p.imax)`?

eschnett:
The mask is set via `mask = i < n`. See the function `mask_for_loop_tail` in the file `simd.hxx` in thorn `Arith`, and also the [NSIMD tutorial](https://agenium-scale.github.io/nsimd/tutorial.html).

If you do not use a mask, then the vector load might access memory outside the array bounds which is undefined behaviour and might segfault.

PR #197, opened 2023-08-07
Derivs: Increase imax by the stencil radius in mask for derivatives

lwJi:


eschnett:
Can you add some comments to the code?

lwJi:
I have added some comments for `mask1`

eschnett:
Are you using these methods for a physics thorn? If so, can you add a test case there so that this thorn won't break in the future?

lwJi:
I have a thorn `TestDerivs` that I haven't pushed it. I will pushed it when everything works and add a unit test to it. The physics thorn is still using their own `derivs.hxx` for the moment. I will try to modify them to call the functions from `Derivs/src/derivs.hxx` when I'm pretty sure everything is working here.

eschnett:
You could add the test thorn to this pull request, and also modify `scripts/carpetx.th` to include `TestDerivs` so that it gets tested as well.

lwJi:
I will do that

PR #208, opened 2023-08-18
CarpetX: Add function ArrayGroupSizeB in driver.cxx

lwJi:


CarpetX/src/driver.cxx
```diff
@@ -1669,14 +1673,15 @@ void SetupGlobals() {
...
+    const size_t factor = group.vartype == CCTK_VARIABLE_COMPLEX ? 2 : 1;
     const nan_handling_t nan_handling = arraygroupdata.do_checkpoint
                                             ? nan_handling_t::forbid_nans
                                             : nan_handling_t::allow_nans;
     arraygroupdata.data.resize(group.numtimelevels);
     arraygroupdata.valid.resize(group.numtimelevels);
     for (int tl = 0; tl < int(arraygroupdata.data.size()); ++tl) {
       // TODO: Allocate in managed memory
-      arraygroupdata.data.at(tl).resize(arraygroupdata.numvars *
+      arraygroupdata.data.at(tl).resize(factor * arraygroupdata.numvars *
```
eschnett, line 1684: I think we need to take `factor` into account also when calculating the pointers to the individual variables.

CarpetX/src/driver.cxx
```diff
@@ -2575,6 +2581,32 @@ CCTK_INT CarpetX_GetBoundarySizesAndTypes(
   return 0;
 }
 
+const int *ArrayGroupSizeB(const cGH *cctkGH, int dir, int gi,
+                           const char *groupname) {
+  if (groupname) {
+    gi = CCTK_GroupIndex(groupname);
+  }
+  assert(gi >= 0 and gi < CCTK_NumGroups());
```
eschnett, line 2589: We're now using `&&` instead of `and`...

CarpetX/src/driver.cxx
```diff
@@ -2575,6 +2581,32 @@ CCTK_INT CarpetX_GetBoundarySizesAndTypes(
...
+
+  // TODO: handle case of no storage allocated
+
+  cGroup group;
+  int ierr = CCTK_GroupData(gi, &group);
+  assert(!ierr);
+  assert(dir >= 0 && dir < group.dim);
+  if (group.grouptype == CCTK_GF) {
```
eschnett, line 2597: I have a slight preference for a `switch` statement here.

eschnett (changes requested):

CarpetX/src/driver.cxx
lwJi: Done

CarpetX/src/driver.cxx
lwJi: Done

CarpetX/src/driver.cxx
lwJi: Where did we calculate the pointers? ( I modified it a bit base on my guessing :) )

PR #227, opened 2023-09-14
TestLoopX: Rename from TestLoop

eschnett:


chcheng3:
Looks like the test failed after the thorn was renamed. The output it was expecting was `testloopx-testloop_gf.it000000.x.tsv` rather than `testloop-testloop_gf.it000000.x.tsv` and similar for y and z. It should be enough to just rename the files in `TestLoopX/test/testloop_outermost_interior`.

eschnett:
Thanks!

PR #230, opened 2023-09-15
HydroBaseX: Add RHS quantities for the Avec variables

jaykalinani:
Since Avec is an evolved variable, we require its RHS grid functions. More importantly, the Avec variables should depend on the these RHS quantities. Thus, the TAGS for these variables have been updated accordingly.

### DIFF
--- a/HydroBaseX/interface.ccl
+++ b/HydroBaseX/interface.ccl
@@ -24,6 +24,10 @@ CCTK_REAL Ye TYPE=gf CENTERING={ccc} TAGS='checkpoint="no"' "electron fraction Y
 
 CCTK_REAL Bvec TYPE=gf CENTERING={ccc} TAGS='parities={+1 -1 -1   -1 +1 -1   -1 -1 +1} checkpoint="no"' { Bvecx Bvecy Bvecz } "Magnetic field B^i"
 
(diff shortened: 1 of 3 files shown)


jaykalinani:
This is what I earlier thought too. I tried with `Avec` of HydroBaseX and `Avec_rhs` in AsterX. This gave us spurious results in Balsara tests, and even the CI tests failed (see, for instance, the recent commits here: https://github.com/jaykalinani/AsterX/compare/main...c2p). Defining the `Avec_rhs` in HydroBaseX, and importantly, adding the correct dependencies (via TAGS) to the Avec, fixes the evolution.

eschnett:
The dependencies will just invalidate variables, they should not be necessary for correctness.

I don't see how moving the definition of `Avec_rhs` could change things. Nothing should depend on where the grid function is defined.

It is difficult to declare the RHS variable for `Avec` if you want to use the `Avec` variable defined in `HydroBase`. You would have to modify the grid function group tags table manually after startup. However, if you declare this RHS relation in HydroBase, then each thorn using HydroBase will be forced to evolve `Avec`.

I would, instead, keep both `Avec` and its RHS as "conserved" variable in `AsterX`, different from the "primitive" `Avec` in HydroBase.

jaykalinani:
I see. It's clear to me now. Thanks!

PR #237, opened 2023-10-05
CarpetX: Redesign ENO stencils

eschnett:
Make the ENO acceptance criterion per-group instead of per-function. Rename stencils for consistency.
Add a few more stencil orders.


rhaas80:
Is still something we want? If so, given the many lines changes (>1000) an explanation what changes would be nice to have to help a reviewer. Ideally, if there is a "mechanical" part to this (renaming functions via search & replace) and a "creative" part then each might be reviewed more quickly separately.

eschnett:
I think this is something we want, I now think the previous implementation is wrong. I was waiting mostly for feedback from Swapnil to see whether this works.

Most of the changes were boilerplate, breaking the main loop over components (grid functions) into two sequential loops and combining all the decisions for each grid function.

This code definitively needs a clean-up. I recently saw a video where an experience C++ programmer stated that each function should fit into 30 lines... Almost there!

PR #242, opened 2023-10-23
Moving boxes example thorn

lucass-carneiro:
This PR provides a new thorn called `MovingBoxesToy` that sets up two refinement boxes and makes them move around in a circle. It is intended as a guide for people wanting to implement more complex moving boxes.

In order to make this thorn possible, I had to change the parameters in `BoxInBox` to `RESTRICTED`


eschnett:
Thanks! This thorn is a good idea.

I don't think that `BoxInBox`'s parameters should be "restricted", i.e. should be visible outside the thorn. If `MovingBoxToy` wants to move the box around then it should decide where the origin of the circle is. Using the position from `BoxInBox` is probably just confusing. Another reason is that these parameters are only used to initialize grid scalars, and looking at these parameters later on will give wrong information.

lucass-carneiro:
Thanks for the feedback! I have removed the `RESTRICTED` from `BoxInBox` and changed the toy. The boxes now move without requiring parameters from `BoxInBox`

eschnett:
You are currently not running the test cases in CI since you are not updating the thorn list in the `scripts` directory.

lucass-carneiro:
I have updated the thorn list inside `scripts`

eschnett:
There are no test cases...

MovingBoxToy/test/circle/performance.yaml
eschnett, line 1: Don't add this file, its contents are not stable.

MovingBoxToy/test/circle.par
```diff
@@ -0,0 +1,59 @@
...
+BoxInBox::radius_z_2[1] = 0.25
+
+CarpetX::boundary_x = "dirichlet"
+CarpetX::boundary_y = "dirichlet"
+CarpetX::boundary_z = "dirichlet"
+CarpetX::boundary_upper_x = "dirichlet"
+CarpetX::boundary_upper_y = "dirichlet"
+CarpetX::boundary_upper_z = "dirichlet"
+
+Cactus::cctk_itlast = 20
```
eschnett, line 44: 20 iterations is a lot. Would 5 iterations suffice? That would make this test case 4 times smaller.

eschnett (changes requested):

lucass-carneiro:
I've removed the `yml` file and reduce the size of the test

eschnett (approved):

PR #251, opened 2023-11-30
CarpetX: Correct openPMD mesh names for multipatch I/O

eschnett:


lucass-carneiro:
I report that this PR seems to work on CPUs.
Now one `openPMD` mesh is produced per patch

lucass-carneiro:
Please see PR #253

PR #255, opened 2023-12-15
Added documentation on the  `parities` tag

lucass-carneiro:


CarpetX/doc/ccl_files.tex
```diff
@@ -154,4 +154,38 @@ \subsubsection{\texttt{checkpoint}}
...
+The \texttt{parity} tag controls a tensor's reflection symmetries about the \texttt{x}, \texttt{y} and \texttt{z} axes. There are 3 parities for each tensor component, representing the symmetries about the 3 spatial axes.
+
+For example, a scalar, has \texttt{parities=\{+1 +1 +1\}}, which indicates that no sign changes take place when the quantity is reflected around the axes. If no \texttt{parities} tag is present in a variable group declaration, it is assumed to be a scalar and \texttt{parities=\{+1 +1 +1\}} is implicitly assumed. For a vector, the \texttt{parities=\{-1+1+1 +1-1+1 +1+1-1\}} tag indicates in the first triad of numbers that the vector's \texttt{x} component changes sign while being reflected in the \texttt{x} direction, the second triad indicates that the \texttt{y} component changes sign while reflected in the \texttt{y} direction and the third triad indicates that the \texttt{z} component changes sign while being reflected in the \texttt{z} direction. As an example, let us look at the parities of the energy momentum tensor in general relativity as declared in the \texttt{interface.ccl} file of the \texttt{TmunuBaseX} thorn included in \CarpetX\space:
```
eschnett, line 162: I don't think the parity numbers can be written without spaces. Can you add spaces here? Maybe also use `\hspace` or somesuch to increase the space between the 3-number-groups.

eschnett (changes requested):

CarpetX/doc/ccl_files.tex
lucass-carneiro: I have Added spaces in the number triplets and wider spaces separating each triplet. In `\texttt{}` mode, I'm using `\hspace` but that does not work in the `lstlisting` environment I'm using for code samples, so in there there are 4 spaces separating each triplet. Visually, it the separation becomes clear in both cases.

eschnett: For scalars the spacing should just be
```
{+1 +1 +1}
```
instead of
```
{+1\hspace{15pt}+1\hspace{15pt}+1\}
```

There is one group of 3 numbers.

lucass-carneiro: Right! Fixed it.

eschnett (approved):

PR #276, opened 2024-04-26
Loop: Add 2 more interior boundary points

shankar-1729:
Required for the Newrad implementation that uses 3 innermost boundary points


eschnett:
Instead of adding more layers of boundary points, can we calculate the distance and direction to the nearest boundary? This would make it straightforward to calculate the information contained in `BI*` in the loop.

PR #277, opened 2024-04-26
max_grid_size as an array

shankar-1729:
Various changes related to setting `max_grid_size` for each level if required, as well as steering `max_grid_size` during evolution.


eschnett:
Can you add a test case for this?

PR #293, opened 2024-06-27
make arrays and scalar accessible from GPU

rhaas80:
These commits use AMReX's arena to allocated storage for grid scalars and grid arrays so that they  are visible to GPUs

### DIFF
--- a/CarpetX/src/driver.cxx
+++ b/CarpetX/src/driver.cxx
@@ -1684,7 +1684,6 @@ void SetupGlobals() {
     arraygroupdata.data.resize(group.numtimelevels);
     arraygroupdata.valid.resize(group.numtimelevels);
     for (int tl = 0; tl < int(arraygroupdata.data.size()); ++tl) {
-      // TODO: Allocate in managed memory
       arraygroupdata.data.at(tl).resize(
           amrex::Box(amrex::IntVect(arraygroupdata.lbnd),
                      amrex::IntVect(arraygroupdata.ubnd)),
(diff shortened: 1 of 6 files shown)


CarpetX/src/driver.cxx
```diff
@@ -2167,7 +2168,14 @@ operator<<(YAML::Emitter &yaml,
   yaml << YAML::BeginMap;
   yaml << YAML::Key << "commongroupdata" << YAML::Value
        << (GHExt::CommonGroupData)arraygroupdata;
-  yaml << YAML::Key << "data" << YAML::Value << arraygroupdata.data;
+  yaml << YAML::Key << "data" << YAML::Value << YAML::Flow << YAML::BeginSeq;
+  for (int tl = 0; tl < int(arraygroupdata.data.size()); ++tl) {
+    yaml << YAML::BeginSeq;
+    for (int vi = 0; vi < arraygroupdata.numvars; ++vi)
+      yaml << *arraygroupdata.data.at(tl).dataPtr(vi);
```
eschnett, line 2175: This seems to output a pointer?

CarpetX/src/driver.hxx
```diff
@@ -14,6 +14,7 @@
 #include <AMReX_FluxRegister.H>
 #include <AMReX_Interpolater.H>
 #include <AMReX_MultiFab.H>
+#include <AMReX_FArrayBox.H>
```
eschnett, line 17: Please run `clang-format` to sort the included files.

eschnett (changes requested):

CarpetX/src/driver.cxx
rhaas80: True, I'll update so that the data output is the same (kind of) data as is output for grid functions.

CarpetX/src/driver.hxx
rhaas80: Will do. Is there a specific version that I should use?

PR #300, opened 2024-07-29
Loop: Check for symmetry boundaries in outermost interior loops

chcheng3:
The `loop_outermost_int` and `loop_outermost_int_device` routines used in NewRadX should check for symmetry boundaries. This is so that at symmetry boundaries, the radiative boundary conditions are applied if and only if they are also adjacent to physical boundaries.

This pull request also updates TestLoopX with the new loop syntax, so that symmetry information is passed into the loop from the patch data. The test case is also updated with reflection symmetry and periodic boundary conditions enabled.


rhaas80:
Ticket is here: https://github.com/EinsteinToolkit/CarpetX/pull/300

lucass-carneiro (changes requested):
Overall the patch appears to be well written. It contains test cases demonstrating its effectiveness.

The patch changes the `loop_outermost_int` API, which may break existing codes that are using that loop construct. Since to my knowledge, this is used only in boundary codes such as NewRadX, this should be OK.

I have very minor comments on adjustments, after which I would recommend for this PR to be merged

TestLoopX/src/testloop.cxx
```diff
@@ -1,4 +1,5 @@
 #include <cctk.h>
+#include "CarpetX/CarpetX/src/driver.hxx"
```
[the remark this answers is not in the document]
chcheng3: I agree with your suggestions, thanks for pointing them out. In order to use the `USES INCLUDE` mechanism for `driver.hxx`, this would require adding a `INCLUDES HEADER: driver.hxx IN driver.hxx` to the `interface.ccl` for CarpetX.

TestLoopX/test/testloopx_outermost_interior.par
```diff
@@ -14,19 +14,39 @@ CarpetX::poison_undefined_values = "yes"
 
 CarpetX::xmin = -10
 CarpetX::ymin = -10
-CarpetX::zmin = -10
+CarpetX::zmin = 0
 CarpetX::xmax = +10
 CarpetX::ymax = +10
 CarpetX::zmax = +10
 
+CarpetX::blocking_factor_x = 2
```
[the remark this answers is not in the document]
chcheng3: Here, I had set `blocking_factor` to 2 in order to make the number of cells divisible by the blocking factor.  However, I agree with you that the test parameter files should be kept simple.

So instead, I am proposing to change `ncells` to be divisible by the default blocking factor (8). I am also changing the grid extent so that the same spacing is kept the same as before (0.5), not that it matters for this test but purely for aesthetic reasons.

CarpetX/interface.ccl
```diff
@@ -14,6 +14,7 @@ USES INCLUDE HEADER: vect.hxx
 USES INCLUDE HEADER: loop.hxx
 USES INCLUDE HEADER: loop_device.hxx
 
+INCLUDES HEADER: driver.hxx IN driver.hxx
```
rhaas80, line 17: This change should be in a commit `CarpetX:`  and not `Loop`.

rhaas80 (changes requested):
please use one commit per thorn that you modify, eg if modifying CarpetX the commit message should start with `CarpetX:`

CarpetX/interface.ccl
eschnett: You cannot include `CarpetX` files from `Loop` because `Carpetx` depends on `Loop`, and this would create a circular dependency.

rhaas80: huh. I am speaking out of turn, I guess, yet: Uh, if Loop does not use the functionality from `driver.hxx` the I think I should not say `uses include header: driver.hxx`. By the same reasoning if the client thorns (say `NewRadX` ) directly use the information then _they_ should contain the `uses include header: driver.hxx` statement.

rhaas80: My take is that circularity is an issue for the `REQUIRE:` statements (they affect compilation order so circles are not allowed) but not the `USES INCLUDE HEADER` statements I would say (handled before dependency is even known to Cactus).  
  
I agree with Erik's concern wrt CarpetX / Loop depending on each other in https://github.com/EinsteinToolkit/CarpetX/pull/300#discussion_r1821138682  (since we'd want a `REQUIRE` to go along with the `uses include header`).

lucass-carneiro (approved):

chcheng3:
@rhaas80 To create a separate commit for the change to `CarpetX`, I reverted to the commit `a3b4594` and separated out the commits to `CarpetX` and `TestLoopX`.

TestLoopX/src/testloop.cxx
```diff
@@ -1,9 +1,9 @@
 #include <cctk.h>
-#include "CarpetX/CarpetX/src/driver.hxx"
 #include <cctk_Arguments.h>
 #include <cctk_Parameters.h>
 #include <loop.hxx>
 #include <loop_device.hxx>
+#include <driver.hxx>
```
rhaas80, line 6: I'd use the double quote form of include using `"` instead of hte `<>` form since none of these files are system include files.

See 

https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

for the details (there won't be a difference in Cactus using gcc since we use `-I` to add the extra paths). 

The C99 standard say (in https://en.cppreference.com/w/cpp/preprocessor/include) that the double quote variant

> The intent of syntax (2) is to search for the files that are not controlled by the implementation.

ie include files not provided by the compiler.

Exactly where you put that cut-off between `"` and `<>` is a matter of some choice it seems.

Loop/src/loop.hxx
```diff
@@ -541,6 +541,7 @@ public:
   template <int CI, int CJ, int CK, int VS = 1, int N = 1, typename F>
   inline CCTK_ATTRIBUTE_ALWAYS_INLINE void
   loop_outermost_int(const vect<int, dim> &group_nghostzones,
+                     const vect<vect<bool, dim>, 2> &is_sym_bnd,
```
rhaas80, line 544: I'd document what `is_sym_bnd[f][d]` stands for. Does this order (f then d) match what is used eg in `cctk_bbox` (see https://www.einsteintoolkit.org/usersguide/UsersGuide.html#x1-98000C1.6.2). Looking at the Cactus example it does seem to be different since the Cactus code uses:

```
An array of 2cctk_dim integers (in the order [dim_0^{min}, dim_0^{max}, dim_1^{min}, dim_1^{max}, ... ]), ...
```

Loop/src/loop_device.hxx
```diff
@@ -275,6 +275,7 @@ public:
             int NT = AMREX_GPU_MAX_THREADS, typename F>
   inline CCTK_ATTRIBUTE_ALWAYS_INLINE void
   loop_outermost_int_device(const vect<int, dim> &group_nghostzones,
+                            const vect<vect<bool, dim>, 2> &is_sym_bnd,
```
rhaas80, line 278: same comment as above.

CarpetX/interface.ccl
rhaas80: Do we _need_ this include in Loop? Since if there is the include I'd normally add a `REQUIRE` (otherwise Cactus may silently give you an empty include file). But the `REQUIRE` will introduce a circular dependency between `Loop` and `CarpetX`

Loop/src/loop.hxx
chcheng3: The `is_sym_bnd` array follows the layout of the `symmetry` array from `driver.hxx`, which seems to be arranged in a different order from what Cactus does:
https://github.com/EinsteinToolkit/CarpetX/blob/a941c336a422030a1d4d72edcc38f69446ee623e/CarpetX/src/driver.hxx#L211
So the patch here is consistent with CarpetX, but not Carpet, which might be confusing.

CarpetX/interface.ccl
chcheng3: Well, what I'm proposing here would _not_ need the include (and hence the `REQUIRE`) in `Loop`. By itself, the `loop_outermost_int` routine does not use any information from the driver. On the other hand, the client thorns that intend to take the symmetry information from the driver and pass it to `loop_outermost_int`, _do_ need the include.

rhaas80: The change still shows up in files changed as adding `INCLUDE HEADER "..."`. So if it is not used by Loop I'd suggest to remove the change.

chcheng3: Hmm, from what I'm seeing the `USES INCLUDE HEADER` is only in `TestLoopX`, but not `Loop`. Can you point me to where it is?

rhaas80: oha, sorry yes you are correct. I was looking at the very wrong file. In fact not even had I confused the thorn but also `INCLUDE HEADER` and `USES INCLUDE HEADER`.

chcheng3:
@eschnett Could you take a moment to review this pull request and let me know your comments? The changes to CarpetX and Loop here are necessary for the NewRadX pull request review (https://github.com/EinsteinToolkit/SpacetimeX/pull/39) to be proceed. Thank you!

eschnett:
Why do you need to know whether a particular boundary is a symmetry boundary? I think what you actually need to know if a particular boundaries is a newrad boundary.

chcheng3:
Yes, we are actually checking whether a certain boundary is a NewRad boundary. This would consist of the points that correspond to the physical outer boundary, but belong to the interior from CarpetX's point of view. If a symmetry is applied to a boundary, then that boundary would not be an outer boundary and not a NewRad boundary. We would need to know which boundaries are symmetry boundaries, so that they can be excluded from the "outermost interior" loops.

eschnett:
Should we wait until after the release branches have been created?

rhaas80:
I am not too worried about this pull request since it only changes C++ code in Loop that so far is only used by NewRadX, though if applying this one also requires the corresponding SpacetiomeX/NewRadX pull request then one needs to be a bit more careful since then the change becomes "user visible".  
  
It is less urgent to include this is in the release now that Z4c has been bumped to the 2025\_05 release then it was when Z4c was to be included in 2024\_11.

eschnett (approved):

lucass-carneiro:
@rhaas80 @eschnett Pushes to Cheng-Hsin's fork don't seem to be triggering the workflow checks. We are also unable to manually trigger them via the `actions` tab because these apply only to branches of this repository and not to forks it seems.

Maybe we are missing something like

```yaml
on:
  pull_request:
    types: [review_requested, edited]
```

in the `CI` file? Otherwise it seems that the checks won't trigger.

rhaas80:
Running for pull requests was disabled in git hash \[e20e3232\](https://github.com/einsteintoolkit/CarpetX/commits/e20e3232b4d172b638672309d5f87b01b0a616c9) "CI: Don't build for pull request changes" of \[CarpetX\](https://github.com/einsteintoolkit/CarpetX) which is about 6 months old.    
  
That would explain the issue. @eschnett is it ok to re-enable?

eschnett:
Yes, that's fine.

I disabled it because I noticed that all my pushes to branches ran CI twice, once via the "push to branch" rule, and once via the "PR" rule. Maybe you know a way to avoid this?

PR #301, opened 2024-07-31
ODESolvers: Solve only in the interior

eschnett:


lwJi:
It breaks unit tests in `WaveToyX` and `TestODESolve`
1. `WaveToyX`: this one can be fixed if we schedule a `WaveToyX_Sync` in `ODESolvers_PostStep`,
2. `TestODESolve`: this one might just because it's not using the loop macros. We might just remove this one from thornlist, because we have another equivalent one called `TestODESolve2` which used loop macros.

rhaas80 (approved):

PR #307, opened 2024-09-08
Extend 'Derivs' to cover 6th and 8th order finite difference

lwJi:
`Derivs`:
1. implement 6th and 8th order accurate central finite difference.
2. implement dissipation and upwind derivatives corresponding to 6th and 8th order central finite difference.

`TestDerivs`:
1. implement unit test for all above. 

all the coefficients used are calculated here using mathematica
https://github.com/lwJi/Docs-Subcycling/blob/main/wolfram/Derivs.ipynb


lwJi:
> Fails due to compile for SYCL due to calls via function pointer:
> 
> https://github.com/EinsteinToolkit/CarpetX/actions/runs/11491459037/job/31983862784?pr=307#step:6:5364
> 
> probably also would fail (at runtime) for CUDA.

I have remove the usage of function pointer, please take another look

Derivs/src/derivs.cxx
```diff
@@ -87,11 +104,51 @@ calc_derivs(const GF3D5<T> &gf, const vec<GF3D5<T>, dim> &dgf,
         });
     break;
 
+  case 8:
+    grid.loop_int_device<CI, CJ, CK, vsize>(
+        grid.nghostzones,
+        [=] CCTK_DEVICE(const PointDesc &p) CCTK_ATTRIBUTE_ALWAYS_INLINE {
+          const vbool mask = mask_for_loop_tail<vbool>(p.i, p.imax);
+          // Take account of ghost points
```
eschnett, line 112: "Take ghost points into account"

Derivs/src/derivs.cxx
```diff
@@ -87,11 +104,51 @@ calc_derivs(const GF3D5<T> &gf, const vec<GF3D5<T>, dim> &dgf,
...
+          const vbool mask1 =
+              mask_for_loop_tail<vbool>(p.i, p.imax + deriv_order / 2);
```
eschnett, line 114: Better to write `8` instead of `deriv_order`. `8` is a constant, `deriv_order` is not.

Derivs/src/derivs.cxx
```diff
@@ -158,84 +215,109 @@ calc_derivs2(const GF3D5<T> &gf, const vec<GF3D5<T>, dim> &dgf,
         });
     break;
 
+  case 8:
+    grid.loop_int_device<CI, CJ, CK, vsize>(
+        grid.nghostzones,
+        [=] CCTK_DEVICE(const PointDesc &p) CCTK_ATTRIBUTE_ALWAYS_INLINE {
+          const vbool mask = mask_for_loop_tail<vbool>(p.i, p.imax);
+          // Take account of ghost points
```
eschnett, line 223: ditto

Derivs/src/derivs.cxx
```diff
@@ -158,84 +215,109 @@ calc_derivs2(const GF3D5<T> &gf, const vec<GF3D5<T>, dim> &dgf,
...
+          const vbool mask1 =
+              mask_for_loop_tail<vbool>(p.i, p.imax + deriv_order / 2);
```
eschnett, line 225: ditto

TestDerivs/interface.ccl
```diff
@@ -20,3 +20,27 @@ CCTK_REAL ddchi TYPE=gf CENTERING={VVV} TAGS='checkpoint="no"'
 {
   dxxchi dxychi dxzchi dyychi dyzchi dzzchi
 } "2nd derivs of test grid function"
+
+CCTK_REAL chi_diss TYPE=gf CENTERING={VVV} TAGS='checkpoint="no"' "dissipation term"
+
+CCTK_REAL beta TYPE=gf CENTERING={VVV}
+{
+  betax, betay, betaz
+}"velocity function used for calculating upwind term"
```
eschnett, line 29: Missing space

TestDerivs/src/test.cxx
```diff
@@ -1,23 +1,89 @@
...
+    coeff = 128;
+    break;
+  case 10:
+    coeff = -512;
+    break;
+  default:
+    assert(0);
+  }
+
+  const int abc = std::abs(coeff);
```
eschnett, line 80: ditto

rhaas80: in fact you may want to use `fabs` to guard against bad include files which only give you `int abs(int)` and not the overloaded ones as well.

lwJi:
Hi @eschnett @rhaas80, I have addressed the comments in the above few commits. Thanks

Derivs/src/derivs.cxx
```diff
@@ -27,6 +26,24 @@ calc_copy(const GF3D5<T> &gf, const GF3D5layout layout,
       });
 }
 
+template <int CI, int CJ, int CK, typename T>
+CCTK_ATTRIBUTE_NOINLINE void
+calc_copy(const vec<GF3D5<T>, dim> &gf, const GF3D5layout layout,
+          const GridDescBaseDevice &grid, const vec<GF3D2<const T>, dim> &gf0) {
+  for (int a = 0; a < dim; ++a)
+    calc_copy<CI, CJ, CK>(gf(a), layout, grid, gf0(a));
+};
```
rhaas80, line 35: This seems to do more than "Extend 'Derivs' to cover 6th and 8th order finite difference" namely it introduces a `calc_copy` function (why this name) for a `vec<GF3D4<T>>` object. 

Nothing in the commit messages gives any indication that this is happening.

Derivs/src/derivs.cxx
```diff
@@ -87,11 +101,50 @@ calc_derivs(const GF3D5<T> &gf, const vec<GF3D5<T>, dim> &dgf,
...
+template <int CI, int CJ, int CK, typename T>
+CCTK_ATTRIBUTE_NOINLINE void
+calc_derivs(const vec<GF3D5<T>, dim> &gf,
+            const vec<vec<GF3D5<T>, dim>, dim> &dgf, const GF3D5layout layout,
+            const GridDescBaseDevice &grid, const vec<GF3D2<const T>, dim> &gf0,
+            const vect<T, dim> dx, const int deriv_order) {
+  for (int a = 0; a < dim; ++a)
+    calc_derivs<CI, CJ, CK>(gf(a), dgf(a), layout, grid, gf0(a), dx,
+                            deriv_order);
+}
```
rhaas80, line 133: these are new as well. Derivs used to only take plain `GF3D5` as inputs.

TestDerivs/param.ccl
```diff
@@ -10,6 +10,7 @@ CCTK_INT deriv_order "Order of spatial finite differencing" STEERABLE=never
   2 :: "Second order finite difference"
   4 :: "Fourth order finite difference"
   6 :: "Sixth order finite difference"
+  8 :: "Eighth order finite difference"
```
rhaas80, line 13: so 6th order was not actually new?

TestDerivs/src/test.cxx
```diff
@@ -49,35 +117,37 @@ extern "C" void TestDerivs_Set(CCTK_ARGUMENTS) {
         const vreal x0 = p.x + Arith::iota<vreal>() * p.dx;
         const CCTK_REAL y0 = p.y;
         const CCTK_REAL z0 = p.z;
-        vreal u0;
-        using std::sin, std::cos;
-        poly(kxx, kxy, kyz, cos(x0), sin(y0), sin(z0), u0);
+        chi.store(mask, p.I,
+                  poly(kxx, kxy, kyz, cos(x0), std::sin(y0), std::sin(z0)));
```
rhaas80, line 121: leftover `std::sin`

TestDerivs/src/test.cxx
```diff
@@ -110,64 +182,170 @@ extern "C" void TestDerivs_CalcDerivs(CCTK_ARGUMENTS) {
...
+          printf("dy = %f\n", dchi(1) - gf_dchi(1)(p.I));
+          printf("dz = %f\n", dchi(2) - gf_dchi(2)(p.I));
+          printf("ddxx = %f\n", ddchi(0, 0) - gf_ddchi(0, 0)(p.I));
+          printf("ddxy = %f\n", ddchi(0, 1) - gf_ddchi(0, 1)(p.I));
+          printf("ddxz = %f\n", ddchi(0, 2) - gf_ddchi(0, 2)(p.I));
+          printf("ddyy = %f\n", ddchi(1, 1) - gf_ddchi(1, 1)(p.I));
+          printf("ddyz = %f\n", ddchi(1, 2) - gf_ddchi(1, 2)(p.I));
+          printf("ddzz = %f\n", ddchi(2, 2) - gf_ddchi(2, 2)(p.I));
 #endif
           assert(0);
```
rhaas80, line 305: don't use `assert(0)` if it's an error use `CCTK_VERROR` or `abort()`. `assert` is turned into a no-op by `NDEBUG`.

TestDerivs/src/test.cxx
```diff
@@ -110,64 +182,170 @@ extern "C" void TestDerivs_CalcDerivs(CCTK_ARGUMENTS) {
...
+        dxychi_error(p.I) = dxychi(p.I) - ddchi(0, 1);
+        dxzchi_error(p.I) = dxzchi(p.I) - ddchi(0, 2);
+        dyychi_error(p.I) = dyychi(p.I) - ddchi(1, 1);
+        dyzchi_error(p.I) = dyzchi(p.I) - ddchi(1, 2);
+        dzzchi_error(p.I) = dzzchi(p.I) - ddchi(2, 2);
+
+        const int diss_order = deriv_order + 2;
+        const CCTK_REAL diss =
+            Arith::pown(-1, diss_order / 2.0 - 1.0) /
+            Arith::pown(2, diss_order) *
```
rhaas80, line 340: since `pown` takes an `int` as its second arguemtn, doing floating point math here seem wrong (since it relies on implicit casts from `double` to `int`) and confusing. Rather use `diss_order / 2 - 1`.

rhaas80 (changes requested):

Derivs/src/derivs.cxx
lwJi: `calc_copy` was there before, what I did is removing the extra argument `dx`, because it was not need.
The reason we need this is because spatial derivatives as stored in `GF3D5`, for consistency, we make a copy of var itself as `GF3D5`

Derivs/src/derivs.cxx
lwJi: this was also there before. What I did is make it more complete and consistent

TestDerivs/param.ccl
lwJi: 6th order upwind and dissipation is new

TestDerivs/src/test.cxx
lwJi: good catch

lwJi:
Hi @rhaas80, I think I have addressed all your comments in the recent few commits. please take another look. Thanks

TestDerivs/src/test.cxx
rhaas80: the change introduces a clang-format-19 change:  
 

```
diff --git a/TestDerivs/src/test.cxx b/TestDerivs/src/test.cxx
index 04ce209b..d919800b 100644
--- a/TestDerivs/src/test.cxx
+++ b/TestDerivs/src/test.cxx
@@ -336,8 +336,7 @@ extern "C" void TestDerivs_CalcError(CCTK_ARGUMENTS) {

         const int diss_order = deriv_order + 2;
         const CCTK_REAL diss =
-            Arith::pown(-1, diss_order / 2 - 1) /
-            Arith::pown(2, diss_order) *
+            Arith::pown(-1, diss_order / 2 - 1) / Arith::pown(2, diss_order) *
             poly_diss(kxx, kxy, kyz, p.x, p.y, p.z, diss_order, dx);
         chi_diss_error(p.I) = diss - chi_diss(p.I);
```

rhaas80:
Formatting change requested. Looks ok otherwise.

lwJi:
> Formatting change requested. Looks ok otherwise.

Just pushed the formatting change. Thanks for catching it

rhaas80 (approved):

rhaas80:
Thank you. Looks good to me. You may want to reduce the (large) number of commits using a round of `git rebase -i origin/main` (makes it a bit easier in the future to see what changes happened).

lwJi:
> Thank you. Looks good to me. You may want to reduce the (large) number of commits using a round of `git rebase -i origin/main` (makes it a bit easier in the future to see what changes happened).

@rhaas80 , good point. I have squash them into 3 comments.

PR #313, opened 2024-10-31
CarpetX: make driver.hxx and schedule.hxx visible to other thorns

lwJi:


lwJi:
`CarpetX::active_levels->loop_serially` and so on

PR #323, opened 2024-12-06
add backtrace functionality from CarpetLib

rhaas80:
This adds the backgrace functionality found in CarpetLib. When a signal is received we walk that stack trace and output address and function name of each function encountered. 

Rank numbers are obtained from `CCTK_MyProc` instead of`dist::rank` as was done in CarpetLib. No other changes really.

There is a parfile in `tests` that triggers a backtrace. However the test cannot be run as part of Cactus' test suite sine the the executable exits with a failure code and also produces no output.

### DIFF
--- a/CarpetX/param.ccl
+++ b/CarpetX/param.ccl
@@ -19,6 +19,10 @@ STRING amrex_parameters[100] "Additional AMReX parameters"
 } ""
 
 
+BOOLEAN test_backtrace "Kill yourself to test the backtrace mechanism"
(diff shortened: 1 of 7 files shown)


rhaas80:
Ticket is here: https://bitbucket.org/einsteintoolkit/tickets/issues/2835/add-backtrace-functionality-from-carpetlib

rhaas80:
@eschnett suggested a separate thorn.

PR #324, opened 2024-12-06
add support for different data types in arrays, also on GPU

rhaas80:
This adds proper support for data types other than `double` to CarpetX, also for GPUs.

It introduces a helper class `AnyVector`that can hold any data type (similar to `gdata` in CarpetLib), as well as a number of switch statements to handle openPMD's typed output routines. 

I am told that very new versions of openPMD-api lets one call the internal un-typed output routine, but at the time this code was written, that was not yet possible, and thus may not be possible on all cluster installed copies of openPMD.

Python code to dump data to compare:

```python
#!/usr/bin/env python3

import openpmd_api as io
import sys
import numpy

series = io.Series(sys.argv[1], io.Access.read_only)

print("Read a Series with openPMD standard version %s" %
      series.openPMD)

print("The Series contains {0} iterations:".format(len(series.iterations)))
for i in series.iterations:
    print("\t {0}".format(i))
print("")

it = 0
i = series.iterations[it]
print("Iteration {0} contains {1} meshes:".format(it, len(i.meshes)))
for m in i.meshes:
    print("\t {0}".format(m))
print("")

gf = i.meshes["testoutput_gf_patch00_lev00"]["testoutput_gf"]
shape = gf.shape
print("Field gf has shape {0} and datatype {1}".format(
      shape, gf.dtype))
arrdata = gf.load_chunk()
series.flush()
numpy.savetxt("testoutput_gf.asc", arrdata.flatten())

for var in ["sc", "a1", "a2", "a3"]:
    for suffix in [""]: #, "_int", "_complex"]:
        data = i.meshes["testoutput_"+var+suffix]["testoutput_"+var+suffix]
        print("Field {0} has shape {1} and datatype {2}".format("testoutput_"+var+suffix, data.shape, data.dtype))
        arrdata = data.load_chunk()
        series.flush()
        numpy.savetxt("testoutput_"+var+suffix+".asc", arrdata.flatten())
```


rhaas80:
Ticket is here: https://bitbucket.org/einsteintoolkit/tickets/issues/2836/add-support-for-different-data-types-to

rhaas80:
Fixing the various CUDA issues right now.

CarpetX/src/io_openpmd.cxx
```diff
@@ -1206,8 +1206,8 @@ void carpetx_openpmd_t::InputOpenPMD(const cGH *const cctkGH,
                 constexpr std::uint64_t ipoison =
                     0xfff8000000000000ULL + 0xdeadbeef;
 #endif
-                char poison[CCTK_VarTypeSize(cgroup.vartype)];
-                std::memcpy(&poison, &ipoison, sizeof(poison));
+                std::vector<char> poison(CCTK_VarTypeSize(cgroup.vartype));
+                std::memcpy(poison.data(), &ipoison, poison.size());
```
eschnett, line 1210: This won't work when `poison.size()` is larger than `sizeof ipoison`.

Maybe define a few poison value for various types instead?

rhaas80: Well it'll copy in something but leave some bits unitialized. 

Right now poison is actually a NaN.

`valid.cxx` instead uses:  
 

```

  const size_t typesize = size_t(CCTK_VarTypeSize(group.vartype));
  char poison[typesize];
  // for int: deadbeef for little endian machines
  for (size_t i = 0; i < typesize; i += sizeof poison)
    std::memcpy(&poison[i], &ipoison, std::min(sizeof poison, typesize - i));
```

which should actually fail with nvcc and `-std=c++17` since variable size arrays are not in C++ (and I got a failure b/c of that in `io_openpmd.cxx` ).  
  
Should we use the same logic as in `valid.cxx` also in `io_openpmd.cxx`?

rhaas80: added like this in the current version

CarpetX/src/driver.hxx
```diff
@@ -152,13 +152,85 @@ struct GHExt {
...
+      AnyTypeVector() : _type(-1), _data(nullptr), _count(0){};
+      AnyTypeVector(int type_, size_t count_)
+          : _type(type_),
+            _data(amrex::The_Arena()->alloc(CCTK_VarTypeSize(type_) * count_)),
+            _count(count_){};
+      // Noncopyable for now
+      AnyTypeVector(const AnyTypeVector &) = delete;
+      AnyTypeVector &operator=(const AnyTypeVector &) = delete;
+      AnyTypeVector &operator=(const AnyTypeVector &&other) {
+        this->_type = other._type;
```
eschnett, line 171: This is incorrect. The moved-from object will still have its destructor called, which would deallocate the memory. The moved-from object needs to be put into an "inactive" state.

CarpetX/src/driver.hxx
```diff
@@ -152,13 +152,85 @@ struct GHExt {
...
+        this->_data = other._data;
+        this->_count = other._count;
+        return *this;
+      }
+      AnyTypeVector(AnyTypeVector &&other)
+          : _type(other._type), _data(other._data), _count(other._count) {}
```
eschnett, line 177: Ditto

CarpetX/src/driver.hxx
```diff
@@ -152,13 +152,85 @@ struct GHExt {
...
+    private:
+      int _type;
+      void *_data;
+      size_t _count;
+
+    public:
+      AnyTypeVector() : _type(-1), _data(nullptr), _count(0){};
+      AnyTypeVector(int type_, size_t count_)
+          : _type(type_),
+            _data(amrex::The_Arena()->alloc(CCTK_VarTypeSize(type_) * count_)),
```
eschnett, line 165: This looks dangerous. Passing an invalid `type` makes `CCTK_VarTypeSize` return a negative number, and C++ will convert this into a very large unsigned number for `alloc`. Better to check and call `CCTK_ERROR`.

CarpetX/src/driver.hxx
```diff
@@ -152,13 +152,85 @@ struct GHExt {
...
+        assert(_data != nullptr);
+        amrex::The_Arena()->free(_data);
+        _type = -1;
+        _data = nullptr;
+        _count = 0;
+      }
+
+      int type() const { return _type; };
+
+      void *data_at(size_t i) const {
```
eschnett, line 218: There should probably be two versions of this function, one `const` and returning a `const void*`, and the other non-const and returning a `void *`.

CarpetX/src/io_openpmd.cxx
```diff
@@ -1155,38 +1154,50 @@ void carpetx_openpmd_t::InputOpenPMD(const cGH *const cctkGH,
...
+            case CCTK_VARIABLE_INT:
+              record_components.at(vi).loadChunkRaw(
+                  static_cast<CCTK_INT *>(cactus_var_ptr), start, count);
+              break;
+            case CCTK_VARIABLE_COMPLEX:
+              record_components.at(vi).loadChunkRaw(
+                  static_cast<CCTK_COMPLEX *>(cactus_var_ptr), start, count);
+              break;
+            default:
+              assert(0 && "Unexpected variable type");
```
eschnett, line 1178: I'd call `CCTK_ERROR` here because it can't be disabled by compiler options.

CarpetX/src/io_openpmd.cxx
```diff
@@ -1195,23 +1206,59 @@ void carpetx_openpmd_t::InputOpenPMD(const cGH *const cctkGH,
                 constexpr std::uint64_t ipoison =
                     0xfff8000000000000ULL + 0xdeadbeef;
 #endif
-                CCTK_REAL poison;
-                std::memcpy(&poison, &ipoison, sizeof poison);
+                const size_t typesize = size_t(CCTK_VarTypeSize(cgroup.vartype));
+                std::vector<char> poison(typesize);
+                // for int: deadbeef for little endian machines
+                for (size_t i = 0; i < typesize; i += sizeof poison)
+                  std::memcpy(&poison[i], &ipoison, std::min(poison.size(), typesize - i));
```
eschnett, line 1213: I would define three different poison values here (for int, real, and complex). Using just some of the `real` bits for `int` might give values that aren't very distinctive. I found that large positive values make for good integer poison. Negative values don't work that well because they are more likely to be clipped off.

CarpetX/src/io_openpmd.cxx
```diff
@@ -1195,23 +1206,59 @@ void carpetx_openpmd_t::InputOpenPMD(const cGH *const cctkGH,
...
+                  std::shared_ptr<CCTK_COMPLEX>(
+                      static_cast<CCTK_COMPLEX *>(
+                          cactus_ptr->data_at(contig_offset + cactus_np * vi)),
+                      [=](CCTK_COMPLEX *const ptr) {
+                        expand_box(static_cast<void *const>(ptr));
+                      }),
+                  start, count);
+              break;
+            default:
+              assert(0 && "Unexpected variable type");
```
eschnett, line 1260: I would call `CCTK_ERROR` here.

CarpetX/src/io_openpmd.cxx
```diff
@@ -1686,8 +1732,18 @@ void carpetx_openpmd_t::OutputOpenPMD(const cGH *const cctkGH,
...
+        const openPMD::Datatype datatype = [](int varType) {
+          switch (varType) {
+          case CCTK_VARIABLE_REAL:
+            return openPMD::determineDatatype<CCTK_REAL>();
+          case CCTK_VARIABLE_INT:
+            return openPMD::determineDatatype<CCTK_INT>();
+          case CCTK_VARIABLE_COMPLEX:
+            return openPMD::determineDatatype<CCTK_COMPLEX>();
+          default:
+            assert(0 && "Unexpected varType");
```
eschnett, line 1744: `CCTK_ERROR`

CarpetX/src/io_openpmd.cxx
```diff
@@ -1771,36 +1827,74 @@ void carpetx_openpmd_t::OutputOpenPMD(const cGH *const cctkGH,
...
-            record_components.at(vi).storeChunk(openPMD::shareRaw(ptr), start,
-                                                count);
+#if !OPENPMDAPI_VERSION_GE(0, 15, 0)
+#define storeChunkRaw(ptr, start, count)                                       \
+  storeChunk(openPMD::shareRaw(ptr), start, count)
 #endif
+            switch (cgroup.vartype) {
+            case CCTK_VARIABLE_REAL:
+              record_components.at(vi).storeChunkRaw(
+                  static_cast<CCTK_REAL const *const>(var_ptr), start, count);
```
eschnett, line 1840: This should be `static_cast<CCTK_REAL const *>`, the additional `const` doesn't have any effect since the pointer is an rvalue.

CarpetX/src/io_tsv.cxx
```diff
@@ -152,14 +155,36 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
...
   file << "\n";
 
   // Output data
   file << cctkGH->cctk_iteration << sep << cctkGH->cctk_time;
   const int tl = 0;
   for (int vi = 0; vi < arraygroupdata.numvars; ++vi)
-    file << sep << arraygroupdata.data.at(tl).at(vi);
+    switch (cgroup.vartype) {
+    case CCTK_VARIABLE_REAL:
+      file << sep << *(CCTK_REAL *)arraygroupdata.data.at(tl).data_at(vi);
```
eschnett, line 174: Should be `(const CCTK_REAL *)`

CarpetX/src/driver.cxx
```diff
@@ -2312,6 +2310,37 @@ YAML::Emitter &operator<<(YAML::Emitter &yaml,
...
+};
+
+YAML::Emitter &
+operator<<(YAML::Emitter &yaml,
+           const GHExt::GlobalData::AnyTypeVector &anytypevector) {
+  yaml << YAML::BeginSeq;
+  for (size_t i = 0; i < anytypevector.size(); ++i) {
+    switch (anytypevector.type()) {
+    case CCTK_VARIABLE_COMPLEX:
+      yaml << *(CCTK_COMPLEX *)anytypevector.data_at(i);
```
eschnett, line 2326: This should be `(const CCTK_COMPLEX *)`.

CarpetX/src/io_tsv.cxx
```diff
@@ -193,7 +221,14 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
...
   for (const auto &varname : varnames)
-    file << sep << col++ << ":" << varname;
+    if (cgroup.vartype == CCTK_VARIABLE_REAL ||
+        cgroup.vartype == CCTK_VARIABLE_INT)
+      file << sep << col++ << ":" << varname;
+    else if (cgroup.vartype == CCTK_VARIABLE_COMPLEX) {
+      file << sep << col++ << ":" << varname << ".real";
+      file << sep << col++ << ":" << varname << ".imag";
+    } else
+      assert(0 && "Unexpected variable type");
```
eschnett, line 231: Better call `CCTK_ERROR`

CarpetX/src/io_tsv.cxx
```diff
@@ -209,7 +244,27 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
       file << sep << (dir == out_dir ? i : 0);
     const int tl = 0;
     for (int vi = 0; vi < arraygroupdata.numvars; ++vi)
-      file << sep << arraygroupdata.data.at(tl).at(np * vi + DI[out_dir] * i);
+      switch (cgroup.vartype) {
+      case CCTK_VARIABLE_REAL:
+        file << sep
+             << *(CCTK_REAL *)arraygroupdata.data.at(tl).data_at(
```
eschnett, line 250: `(const CCTK_REAL *)`

CarpetX/src/io_tsv.cxx
```diff
@@ -209,7 +244,27 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
...
+                    np * vi + DI[out_dir] * i);
+        break;
+      case CCTK_VARIABLE_INT:
+        file << sep
+             << *(CCTK_INT *)arraygroupdata.data.at(tl).data_at(
```
eschnett, line 255: The expression `arraygroupdata.data.at(tl).data_at(np * vi + DI[out_dir] * i)` could be calculated before the `switch` statement to reduce code duplication.

CarpetX/src/valid.cxx
```diff
@@ -229,24 +229,28 @@ void poison_invalid_ga(const int gi, const int vi, const int tl) {
...
+  assert(!ierr);
 
   const valid_t &valid = arraygroupdata.valid.at(tl).at(vi).get();
   if (valid.valid_all())
     return;
 
-  CCTK_REAL poison;
-  std::memcpy(&poison, &ipoison, sizeof poison);
+  const size_t typesize = size_t(CCTK_VarTypeSize(group.vartype));
+  char poison[typesize];
```
eschnett, line 241: Didn't you switch to `std::vector` elsewhere for a similar construct?

CarpetX/src/valid.cxx
```diff
@@ -477,33 +481,50 @@ void check_valid_ga(const int gi, const int vi, const int tl,
...
   std::size_t nan_count{0};
 
   const valid_t &valid = arraygroupdata.valid.at(tl).at(vi).get();
   if (valid.valid_all())
     return;
 
-  CCTK_REAL poison;
-  std::memcpy(&poison, &ipoison, sizeof poison);
+  const size_t typesize = size_t(CCTK_VarTypeSize(group.vartype));
+  char poison[typesize];
```
eschnett, line 495: `std::vector`?

CarpetX/src/valid.cxx
```diff
@@ -477,33 +481,50 @@ void check_valid_ga(const int gi, const int vi, const int tl,
...
-                              false))
-        ++nan_count;
+    if (group.vartype == CCTK_VARIABLE_REAL) {
+      const CCTK_REAL *restrict const ptr = static_cast<const CCTK_REAL *const>(
+          arraygroupdata.data.at(tl).data_at(vi));
+      for (int i = 0; i < n_elems; i++) {
+        using std::isnan;
+        if (CCTK_BUILTIN_EXPECT(nan_handling == nan_handling_t::allow_nans
+                                    ? std::memcmp(ptr, poison, sizeof *ptr) == 0
+                                    : isnan(*ptr),
```
eschnett, line 516: You could call `isnan(real) || isnan(imag)` for complex numbers.

eschnett (changes requested):
I've left a few comments. Some are important and need to be addressed. Most are suggestions.

CarpetX/src/valid.cxx
rhaas80: Yes, and *here* it dies not complain about the varuable sized array, but it diez in io_openpmd.cxx. I have no clue why.

I will redo thus all anyway, to have one poision code for bith places where it is used.

CarpetX/src/valid.cxx
rhaas80: I guess, yes.

CarpetX/src/driver.hxx
rhaas80: Fixed in 617d0d9

CarpetX/src/driver.hxx
rhaas80: Fixed in 617d0d9

CarpetX/src/driver.hxx
rhaas80: Moved into constructor code  and / or added an `assert` for the correct values. Those who compile with `NDEBUG` are on their own. In 4fc6ceb

CarpetX/src/driver.hxx
rhaas80: Done in 3605564. Note that differs from AMReX's `Array4D` and \`GridPtrDesc1\` class (and thus requires a const-cast somehwere in schedule.cxx)  
  
For grid-functions:  
\`\`\`  
    auto &restrict groupdata = \*leveldata.groupdata.at(gi);  
   const GridPtrDesc1 grid1(leveldata, groupdata, mfp);  
   for (int tl = 0; tl \< int(groupdata.mfab.size()); ++tl) {  
     const amrex::Array4\<CCTK\_REAL> vars =  
         groupdata.mfab.at(tl)->array(mfp.index());  
     for (int vi = 0; vi \< groupdata.numvars; ++vi)  
       cctkGH->data\[groupdata.firstvarindex + vi\]\[tl\] = grid1.ptr(vars, vi);  
   }  
\`\`\`  
  
For grid-arrays (and scalars):  
\`\`\`  
        auto &restrict arraygroupdata = \*globaldata.arraygroupdata.at(gi);  
       for (int tl = 0; tl \< int(arraygroupdata.data.size()); ++tl) {  
         const auto &restrict vars = arraygroupdata.data.at(tl);  
         for (int vi = 0; vi \< arraygroupdata.numvars; ++vi)  
           cctkGH->data\[arraygroupdata.firstvarindex + vi\]\[tl\] =  
               const\_cast\<void\*>(vars.data\_at(vi \* arraygroupdata.array\_size));  
\`\`\`

CarpetX/src/io_openpmd.cxx
rhaas80: this is not user triggerable (via parfile input) and the list is fixed by Cactus (not user or CarpetX code), so I'd rather keep the assert, if possible.

CarpetX/src/io_openpmd.cxx
rhaas80: done in 49de23b

CarpetX/src/io_openpmd.cxx
rhaas80: https://github.com/EinsteinToolkit/CarpetX/pull/324#discussion_r1899174351

CarpetX/src/io_openpmd.cxx
rhaas80: https://github.com/EinsteinToolkit/CarpetX/pull/324#discussion_r1899174351

CarpetX/src/io_openpmd.cxx
rhaas80: Done in 016fec7

CarpetX/src/io_tsv.cxx
rhaas80: Done in 4704af8

CarpetX/src/driver.cxx
rhaas80: Done in 8f8fbad

CarpetX/src/io_tsv.cxx
rhaas80: Since this is not user (parfile or other runtime input) triggerable and the list of allowed options is determined by the flesh and not the thorn being written (CarpetX) I'd rather keep the assert (since it's a bug) rather than the `CCTK_Error`.

CarpetX/src/io_tsv.cxx
rhaas80: Done in [4704af8](https://github.com/EinsteinToolkit/CarpetX/commit/4704af84e360169589e61c330816d0dd9e10f79a)

CarpetX/src/valid.cxx
rhaas80: removed in 1834168

CarpetX/src/valid.cxx
rhaas80: done in 49de23b

CarpetX/src/io_tsv.cxx
rhaas80: It could, something like this:  
\`\`\`  
    for (int vi = 0; vi \< arraygroupdata.numvars; ++vi)  
     file \<\< sep;  
     const void \*ptr = arraygroupdata.data.at(tl).data\_at(np \* vi + DI\[out\_dir\] \* i);  
     switch (cgroup.vartype) {  
     case CCTK\_VARIABLE\_REAL:  
       file \<\< \*(const CCTK\_REAL \*)ptr;  
       break;  
     case CCTK\_VARIABLE\_INT:  
       file \<\< \*(const CCTK\_INT \*)ptr;  
       break;  
     case CCTK\_VARIABLE\_COMPLEX: {  
       CCTK\_COMPLEX value = \*(const CCTK\_COMPLEX \*)ptr;  
       file \<\< value.real() \<\< sep \<\< value.imag();  
     } break;  
     default:  
       assert(0 && "Unexpected variable type");  
       break;  
     }  
   file \<\< "\\n";  
 }  
\`\`\`  
though it comes at the price of information about what is output being less local and requiring backtracking in the code when reading it.

eschnett: In addition to `AnyVector`, you could introduce an `AnyScalar`, and add a virtual function that outputs to a stream. (Not sure whether that makes the code simpler or more complex in the end.)

rhaas80: Hmm, maybe. Not having all those `switch` scattered around would be nice.   
  
The tricky one is the `CCTK_COMPLEX` output since it currently uses `sep` which would not be known to `AnyScalar::operator<<` .  
  
Still an `AnyScalar` and a `AnyScalar AnyVector::operator[](size_t i) const` might be useful. Would be a return by value (at least unless I want to really complicate my life and have the AnyScalar carry a back-reference to its parent `AnyVector` so that assignment to it would change the `AnyVector` element) for now.

rhaas80: Something like 98ab37c ? So far only used for output. Could possibly also be used for setting poison values (though there performance issues may show up).

rhaas80:
ping

rhaas80:
anyone interested in this?

eschnett (approved):

PR #326, opened 2024-12-23
CarpetX: Add interpolator self-checks

eschnett:


eschnett:
This mostly improves error messages (when built with `CCTK_DEBUG`).

CarpetX/src/interpolate.cxx
```diff
@@ -544,8 +570,8 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
              ++pti) {
           const auto &particles = pti.GetArrayOfStructs();
           const int component = MFPointer(pti).index();
-          CCTK_VINFO("patch %d level %d component %d old_nparticles: %zu",
-                     patch, lev, component, particles.size());
+          // CCTK_VINFO("patch %d level %d component %d old_nparticles: %zu",
+          //            patch, lev, component, particles.size());
```
rhaas80, line 574: I'd remove or move behind a #ifdef DEBUG this code.

CarpetX/src/interpolate.cxx
```diff
@@ -578,8 +604,8 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
              ++pti) {
           const int component = MFPointer(pti).index();
           const auto &particles = pti.GetArrayOfStructs();
-          CCTK_VINFO("patch %d level %d component %d new_nparticles: %zu",
-                     patch, lev, component, particles.size());
+          // CCTK_VINFO("patch %d level %d component %d new_nparticles: %zu",
+          //            patch, lev, component, particles.size());
```
rhaas80, line 608: I'd remove or move behind a #ifdef DEBUG this code.

rhaas80: (also similar in line 611ff below)

CarpetX/src/interpolate.cxx
```diff
@@ -850,6 +889,7 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
   // Set result
   CCTK_REAL *const restrict *const restrict resultptrs =
       static_cast<CCTK_REAL *const *>(resultptrs_);
+  assert(int(recvbuf.size()) == (nvars + 1) * npoints);
```
rhaas80, line 892: `assert` is too late the code in `CCTK_DEBUG` in line 876 alreay touches all points.

rhaas80: This `assert` can trigger if there are points outside of the domain, can't it?   
  
In that case this `assert` does not adhere to the API set out in the ReferenceManual for `CCTK InterpGridArrays` which documents `CCTK ERROR INTERP POINT OUTSIDE` as a possible return value and does not indicate that the code would abort.

rhaas80 (changes requested):

rhaas80:
@eschnett ping.

PR #328, opened 2025-01-06
CarpetX: CarpetX errors out when it cycles time levels if those levels are not valid everywhere.

stevenrbrandt:
CarpetX: CarpetX errors out when it cycles time levels if those levels are not valid everywhere.

This breaks code using presync-only.

This commit causes CarpetX to sync these grid functions (if presync-only is enabled) instead of erroring out.


CarpetX/src/schedule.cxx
```diff
@@ -1515,12 +1515,30 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
         }
         // All time levels (except the current) must be valid everywhere for
         // checkpointed groups
-        if (groupdata.do_checkpoint)
-          for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl)
-            for (int vi = 0; vi < groupdata.numvars; ++vi)
+        bool presync_only = CCTK_EQUALS(presync_mode, "presync-only");
+        if (presync_only && groupdata.do_checkpoint) {
+          std::set<int> sync_set;
```
eschnett, line 1520: A `std::vector<bool>(CCTK_Num_Groups())` would probably be faster.

CarpetX/src/schedule.cxx
```diff
@@ -1515,12 +1515,30 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
+            }
+          }
+          if (!sync_set.empty()) {
+            std::vector<int> sync_vec(sync_set.begin(), sync_set.end());
+            SyncGroupsByDirI(cctkGH, sync_vec.size(), sync_vec.data(), nullptr);
+          }
+        }
+        // check that what we have is correct
+        if (groupdata.do_checkpoint) {
+          std::set<int> sync_set;
```
eschnett, line 1533: This seems unused.

CarpetX/src/schedule.cxx
```diff
@@ -1515,12 +1515,30 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
         // checkpointed groups
-        if (groupdata.do_checkpoint)
-          for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl)
-            for (int vi = 0; vi < groupdata.numvars; ++vi)
+        bool presync_only = CCTK_EQUALS(presync_mode, "presync-only");
+        if (presync_only && groupdata.do_checkpoint) {
+          std::set<int> sync_set;
+          for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl) {
+            for (int vi = 0; vi < groupdata.numvars; ++vi) {
+              sync_set.insert(gi);
```
eschnett, line 1523: That's weird. You insert the same number into the set several times.

eschnett:
If I understand, this patch adds code before cycling time levels in presync mode. Is that correct? It doesn't change existing code?

CarpetX/src/schedule.cxx
```diff
@@ -1516,26 +1516,18 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
-          std::set<int> sync_set;
           for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl) {
             for (int vi = 0; vi < groupdata.numvars; ++vi) {
-              error_if_invalid(groupdata, vi, tl, make_valid_all(), []() {
-                return "CycleTimelevels for the state vector";
-              });
+              if(presync_only) {
+                std::vector<int> sync_vec;
+                sync_vec.push_back(gi);
+                SyncGroupsByDirI(cctkGH, sync_vec.size(), sync_vec.data(), nullptr);
```
rhaas80, line 1525: that only ever has a single group in `sync_vec` and then will sync that group `groupdata.numvars` times, won't it?

Much better would seem to collect all groups that need syncs in a `for (int gi = 0; gi < num_groups; ++gi) {` first, then sync them all.

CarpetX/src/schedule.cxx
```diff
@@ -1516,26 +1516,18 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
+              } else {
+                error_if_invalid(groupdata, vi, tl, make_valid_all(), []() {
+                  return "CycleTimelevels for the state vector";
```
rhaas80, line 1528: I'd add the group name of the bad group to the output.

rhaas80 (changes requested):

rhaas80:
Ticket is https://bitbucket.org/einsteintoolkit/tickets/issues/2856/carpetx-carpetx-errors-out-when-it-cycles

CarpetX/src/schedule.cxx
```diff
@@ -1567,6 +1566,13 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
     }
 
   } // for gi
+  if(presync_sync_set.size() > 0) {
+    std::vector<int> presync_sync_vec;
+    for(auto i = presync_sync_set.begin(); i != presync_sync_set.end(); ++i) {
```
einsteintoolkit-release, line 1571: i -> gi ?

einsteintoolkit-release: I wouldn't encode the type of `presync_sync_set` in its name (unless this was a Windows (TM) code ...).

CarpetX/src/schedule.cxx
```diff
@@ -1515,14 +1517,11 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
         // checkpointed groups
-        bool presync_only = CCTK_EQUALS(presync_mode, "presync-only");
         if (groupdata.do_checkpoint) {
           for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl) {
             for (int vi = 0; vi < groupdata.numvars; ++vi) {
               if(presync_only) {
-                std::vector<int> sync_vec;
-                sync_vec.push_back(gi);
-                SyncGroupsByDirI(cctkGH, sync_vec.size(), sync_vec.data(), nullptr);
+                presync_sync_set.insert(gi);
```
einsteintoolkit-release, line 1524: this does not handle past timelevels, since `SyncGroupsByDirI` only works on the current `tl==0` timelevel, but the for loop `for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl) {` checks past timelevels (as well)

rhaas80: note that the way the `for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl)` loop is set up the *only* time one would see `tl==0` would be if `ntls==1` ie there is only a single time level (ie no cycling happens since there is nothing to cycle.).

CarpetX/src/schedule.cxx
```diff
@@ -1515,12 +1517,20 @@ void CycleTimelevels(cGH *restrict const cctkGH) {
...
-          for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl)
-            for (int vi = 0; vi < groupdata.numvars; ++vi)
-              error_if_invalid(groupdata, vi, tl, make_valid_all(), []() {
-                return "CycleTimelevels for the state vector";
-              });
+        if (groupdata.do_checkpoint) {
+          for (int tl = (ntls == 1 ? 0 : 1); tl < ntls; ++tl) {
+            for (int vi = 0; vi < groupdata.numvars; ++vi) {
+              // it is only possible to sync time-level zero
+              if(tl == 0 && presync_only) {
```
eschnett, line 1524: You can pull this `if` statement out of the `vi` loop since you're not accessing `vi`.

CarpetX/src/schedule.cxx
eschnett: You can use the new C++ syntax here: `for (auto gi: presync_sync_set)`.

CarpetX/src/schedule.cxx
eschnett: No need to add the group name, it's output automatically when this error message is used.

stevenrbrandt:
AFAIK, I have addressed all requests for change.

CarpetX/src/schedule.cxx
stevenrbrandt: Yes, it only handles the current time level and that is the behavior that I need.

stevenrbrandt:
I have resolved all conversations.

rhaas80 (approved):
Please *squash* all commits and use a correctly formatted commit message

PR #330, opened 2025-01-13
Silo: support filesystem in std::experimental

rhaas80:
CarpetX’s Silo output (and only it) uses std::filesystemwhich is defined in C++17 but not all compilers fully support it, RoCM on Frontier uses std::experimental::filesystem instead.


rhaas80:
Ticket is here: https://bitbucket.org/einsteintoolkit/tickets/issues/2845/support-again-filesystem-in-std

stevenrbrandt (approved):

PR #331, opened 2025-01-16
CarpetX: avoid features not supported by nvhpc's compiler

rhaas80:
This avoids two features not supported by nvhpc version nvc++ 24.3 ("nvc++ 24.3-0 linuxarm64 target on aarch64 Linux -tp neoverse-v2"):

* does not support custom OpenMP reductions
* RANK() is not a F90 function but a GNU extension

The former is the more annoying fix since I added an `#ifdef`. If one consider the possible loss in speed due to using a `#pragma omp critical` instead of (possibly) are more optimal implementation via `#pragma reduction(...)` negligible then one can remove the `#ifdef`.

### DIFF
--- a/CarpetX/src/reduction.cxx
+++ b/CarpetX/src/reduction.cxx
@@ -201,7 +201,14 @@ reduction<CCTK_REAL, dim> reduce(int gi, int vi, int tl) {
       // TODO: check that multi-threading actually helps (and we are
       // not dominated by memory latency)
       // TODO: document required version of OpenMP to use custom reductions
+#ifdef __NVCOMPILER
+#pragma omp parallel
+      {
+      auto& outer(red);
+      reduction<CCTK_REAL, dim> red;
+#else
 #pragma omp parallel reduction(reduction : red)
+#endif
       for (amrex::MFIter mfi(mfab, mfitinfo); mfi.isValid(); ++mfi) {
         const amrex::Box &bx = mfi.tilebox(); // current tile (without ghosts)
         const vect<int, dim> tmin{bx.smallEnd(0), bx.smallEnd(1),
@@ -233,6 +240,11 @@ reduction<CCTK_REAL, dim> reduce(int gi, int vi, int tl) {
         red += reduce_array(vars, vi, tmin, tmax, indextype, imin, imax,
                             finemask.get(), x0, dx);
       }
+#ifdef __NVCOMPILER
+#pragma omp critical
+      outer += red;
+      }
+#endif
     }
   }
 
--- a/TestArrayGroup/src/TestDynamicData.F90
+++ b/TestArrayGroup/src/TestDynamicData.F90
@@ -6,24 +6,21 @@ subroutine TestArrayGroup_DynamicDataF(CCTK_ARGUMENTS)
   DECLARE_CCTK_PARAMETERS
   DECLARE_CCTK_ARGUMENTS
 
+  integer, dimension(3) :: dim3
+
+  ! RANK() is not supported by all compilers, so we fail to complile here instead
   ! Validate grid array dynamic data
-  if(RANK(test1) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test1 array dynamic data")
-  endif
+  dim3 = SHAPE(test1)
   if(SIZE(test1, 1) /= 5 .or. SIZE(test1, 2) /= 6 .or. SIZE(test1, 3) /= 4) then
       call CCTK_ERROR("incorrect size in test1 array dynamic data")
   endif
 
-  if(RANK(test2) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test2 array dynamic data")
-  endif
+  dim3 = SHAPE(test2)
   if(SIZE(test2, 1) /= 5 .or. SIZE(test2, 2) /= 6 .or. SIZE(test2, 3) /= 4) then
       call CCTK_ERROR("incorrect size in test2 array dynamic data")
   endif
 
-  if(RANK(test3) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test3 array dynamic data")
-  endif
+  dim3 = SHAPE(test3)
   if(SIZE(test3, 1) /= 5 .or. SIZE(test3, 2) /= 6 .or. SIZE(test3, 3) /= 4) then
       call CCTK_ERROR("incorrect size in test3 array dynamic data")
   endif
--- a/CarpetX/src/reduction.hxx
+++ b/CarpetX/src/reduction.hxx
@@ -121,7 +121,9 @@ template <typename T, int D> struct combine {
 };
 
 typedef reduction<CCTK_REAL, dim> reduction_CCTK_REAL;
+#ifndef __NVCOMPILER
 #pragma omp declare reduction(reduction:reduction_CCTK_REAL : omp_out += omp_in)
+#endif
 
 MPI_Datatype reduction_mpi_datatype_CCTK_REAL();
 MPI_Op reduction_mpi_op();


lwJi:
I confirm the this change make CarpetX compile on vista with nvc compilers

CarpetX/src/reduction.cxx
```diff
@@ -201,7 +201,14 @@ reduction<CCTK_REAL, dim> reduce(int gi, int vi, int tl) {
       // TODO: check that multi-threading actually helps (and we are
       // not dominated by memory latency)
       // TODO: document required version of OpenMP to use custom reductions
+#ifdef __NVCOMPILER
+#pragma omp parallel
+      {
+      auto& outer(red);
```
eschnett, line 207: Are you sure that this initializes a reference to `red`? I only know the syntax `auto& outer = red`.

TestArrayGroup/src/TestDynamicData.F90
```diff
@@ -6,24 +6,21 @@ subroutine TestArrayGroup_DynamicDataF(CCTK_ARGUMENTS)
   DECLARE_CCTK_PARAMETERS
   DECLARE_CCTK_ARGUMENTS
 
+  integer, dimension(3) :: dim3
+
+  ! RANK() is not supported by all compilers, so we fail to complile here instead
```
eschnett, line 11: `complile`

TestArrayGroup/src/TestDynamicData.F90
```diff
@@ -6,24 +6,21 @@ subroutine TestArrayGroup_DynamicDataF(CCTK_ARGUMENTS)
...
   ! Validate grid array dynamic data
-  if(RANK(test1) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test1 array dynamic data")
-  endif
+  dim3 = SHAPE(test1)
```
eschnett, line 13: `dim3` seems unused. Did you want to test the length of `dim3` as replacement of the rank test?

TestArrayGroup/src/TestDynamicData.F90
eschnett, line 11: No need to add a comment on `RANK` if the code doesn't actually use `RANK`.

TestArrayGroup/src/TestDynamicData.F90
```diff
@@ -6,24 +6,21 @@ subroutine TestArrayGroup_DynamicDataF(CCTK_ARGUMENTS)
...
   if(SIZE(test1, 1) /= 5 .or. SIZE(test1, 2) /= 6 .or. SIZE(test1, 3) /= 4) then
       call CCTK_ERROR("incorrect size in test1 array dynamic data")
   endif
 
-  if(RANK(test2) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test2 array dynamic data")
-  endif
+  dim3 = SHAPE(test2)
```
eschnett, line 18: ditto

TestArrayGroup/src/TestDynamicData.F90
```diff
@@ -6,24 +6,21 @@ subroutine TestArrayGroup_DynamicDataF(CCTK_ARGUMENTS)
...
   if(SIZE(test2, 1) /= 5 .or. SIZE(test2, 2) /= 6 .or. SIZE(test2, 3) /= 4) then
       call CCTK_ERROR("incorrect size in test2 array dynamic data")
   endif
 
-  if(RANK(test3) /= 3) then ! note rank 3 b/c of vector of rank=2 arrays
-      call CCTK_ERROR("incorrect dimension in test3 array dynamic data")
-  endif
+  dim3 = SHAPE(test3)
```
eschnett, line 23: ditto

eschnett (changes requested):
I added comments requesting changes.

CarpetX/src/reduction.cxx
rhaas80: hmm, it compiles, not sure really what else it would do. I'll give it an full test though also making sure it really gives the correct value and code.

TestArrayGroup/src/TestDynamicData.F90
rhaas80: ups.

rhaas80: right.

TestArrayGroup/src/TestDynamicData.F90
rhaas80: the compiler aborts if the rank of dim3 is not correct. So while it is unused the assignment still test that `test1` is of the expected shape. I'll add a comment

CarpetX/src/reduction.cxx
rhaas80: Looks correct to me:  
  
\`\`\`  
Disassembly of section .text:

0000000000000000 \<\_Z3refv>:  
void ref() {  
  0:   55                      push   %rbp  
  1:   48 89 e5                mov    %rsp,%rbp  
 int a;  
 int &b(a);  
  4:   48 8d 45 f4             lea    -0xc(%rbp),%rax  
  8:   48 89 45 f8             mov    %rax,-0x8(%rbp)  
 b = 12;  
  c:   48 8b 45 f8             mov    -0x8(%rbp),%rax  
 10:   c7 00 0c 00 00 00       movl   $0xc,(%rax)  
}  
 16:   90                      nop  
 17:   5d                      pop    %rbp  
 18:   c3                      ret  
\`\`\`  
  
C++ lang docs:  
  
[https://en.cppreference.com/w/cpp/language/reference_initialization](https://en.cppreference.com/w/cpp/language/reference_initialization)  
  
\> Non-list-initialization

<table><tbody><tr><td colspan="10">&nbsp;</td></tr><tr><td><p>T ﻿<strong><code>&amp;</code></strong> ref <strong><code>=</code></strong> target <strong><code>;</code></strong><br>&nbsp;</p><p>T ﻿<strong><code>&amp;</code></strong> ref <strong><code>(</code></strong> target <strong><code>);</code></strong></p></td><td>(1)</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td><td>&nbsp;</td></tr></tbody></table>

it is admittedly not something I have seen a lot (but would happen with this syntax when initializing a member variable. Eg:  
  
\`\`\`  
class foo {  
   const int &ri;  
   foo(int i) : ri(i) {};  
}  
\`\`\`  
  
CarpetX uses the `auto &foo = bar` variant though, so I'll update this to match the remainder of the code.  
  
Thanks!

rhaas80:
Alright, I updated the C++ stuff with `auto &foo = bar` and ran through `clang-format-19` which (predictably) indented lots of code :-(

rhaas80:
@lwJi please click on "approve" if you are happy with the proposed change.

lwJi:
> @lwJi please click on "approve" if you are happy with the proposed change.

I can't find a "approve" button on my side

rhaas80:
@eschnett are you happy with the changes I had implemented? This would still be good to merge in.

CarpetX/src/reduction.cxx
```diff
@@ -201,38 +201,50 @@ reduction<CCTK_REAL, dim> reduce(int gi, int vi, int tl) {
...
+          }
+
+          red += reduce_array(vars, vi, tmin, tmax, indextype, imin, imax,
+                              finemask.get(), x0, dx);
         }
-
-        red += reduce_array(vars, vi, tmin, tmax, indextype, imin, imax,
-                            finemask.get(), x0, dx);
+#ifdef __NVCOMPILER
+#pragma omp critical
```
eschnett, line 244: You can give a name to a critical region. I think this creates a custom lock instead of using the global lock.

eschnett (changes requested):

CarpetX/src/reduction.cxx
rhaas80: right, implemented in 6bb953926642fbbd205bb2106dbb1f755f217233

PR #344, opened 2025-04-11
Derivs: add template instantiation for cell centered grid functions

lwJi:


rhaas80:
Since, as far as I recall, in CarpetX, we tie centering `<1,1,1>` to variable meaning ("is an average") can you explain what the derivative actually does? Computing a derivative of averaged quantities is "strange", since you'd be doing something like:  
  
\`\`\`  
\\frac{\\partial}{\\partial x} \\left( \\int\_{x-h/2}^{x+h/2} \\rho(x') dx' / h \\right) = (rho(x+h/2) - rho(x-h/2)) / h  
\`\`\`  
  
which is "strange" (but yes it _is_ equal to `\partial \rho(x) / \partial x + O(h^2)`.

lwJi:
1. This is only used for the purpose of higher order flux correction.
2. There might be a cell centered version of Z4cow at some point.

rhaas80:
And it actually computes something sensible? Ie what does the `calc_copy<1,1,1>` routine actually compute?

lwJi:
It only used to detect shocks.
`calc_copy` is only copy from GF3D2 to GF3D5

rhaas80:
Ok, so `calc_copy<1,1,1>` seems ok. `calc_deriv<1,1,1>` if it does indeed get used to compute "an approximation to the cell center value of that cell" is not any better than just using the cell average directly which is also equal to the value at the center up to `O(h^2)`, isn't it?

lwJi:
I don't understand. I need cell-centered derivative of a cell-centered grid function for shock detection purpose. The easiest way is to use finite difference. right?

rhaas80:
By "cell centered" what do you mean? CarpetX thinks of "\<1,1,1>" as "an average over the cell", and _not_ "the value of a function \[eg the metric\] sampled at the location of the center of the cell".   
  
Which one do you actually have?

lwJi:
It’s cell-average. But also an approximation of cell center value. Do you have a even easier way to construct derivative than finite difference

rhaas80:
Well, as long as you have checked that indeed `calc_derivs2` is a 2nd order accurate approximation of the value at the center of the the cell derivative of the continuous function that was averaged to compute the cell averaged values, then this would be ok-ish.

How is it computed though? Just as a centered finite difference?  This would not be quite correct since it would differ from what a higher-order accurate derivative would have to do. 

Or (much more correctly) as a 2nd order reconstruction and derivative (which may indeed give the same expression, IDK)? I.e. compute the 2nd order polynomial `\rho = a + b x + c x^2` such that `\int_{xi-h/2}^{xi+h/2} \rho(x') dx' / h = rho_grid(xi)` then the value of `\partial \rho / \partial x` at `xi`.

lwJi:
`Derivs` only provide finite difference

lwJi:
I think cell-centered grid is more fundamental at amrex level. However, the user can use it to represent cell-average (just as what we did with AsterX).

This PR just provides with the user that if they are using cell-centered grid and use it as sampling at that location they can take derivatives using `Derives`.

lwJi:
If the user is using cell-centered grid as cell-average, and finite difference is not an approximation for the derivatives, they can just not call `Derivs` to get the derivatives.

lwJi:
When you say that CarpetX assume that `<1,1,1>` is cell average, I believe you are talking about prolongation. But I believe we have prolongation operator which is not conserved also.

rhaas80:
> When you say that CarpetX assume that `<1,1,1>` is cell average, I believe you are talking about prolongation. But I believe we have prolongation operator which is not conserved also.

only in a `<..,0,...>` direction. CarpetX does currently not handle the "at center but not an average" case.

rhaas80:
hmm, I guess we currently cannot use these since CarpetX will pick the prolongation operator based on the centering. 

Still not quite sure what your cell centered derivatives act on? On a function sampled at the center or a derivative of the continuous function whose cell averages are stored in the cell?

PR #346, opened 2025-06-17
CarpetX: use pinned Particles in CarpetX_Interpolate

lwJi:
The creation of particles is very slow on managed memory. Using pinned memory solve the problem
https://github.com/AMReX-Codes/amrex/discussions/4506


rhaas80:
Would this also work (efficiently) on CPUs?

lwJi:
> Would this also work (efficiently) on CPUs?

I didn't observe any performance difference on CPUs.

rhaas80:
well probably not enough to worry. Worst case would be that two copies of the particle data are kept. So it would be number-of-particles times size-in-bytes-of-each-particle. So probably still (for what we do right now) only "couple thousand particles at most" and "less than a dozen values per particle" so <10000*12*8 = 1MB so basically "we don't care".

rhaas80 (approved):

PR #350, opened 2025-09-26
ADMBaseX: declare use of loop.hxx

rhaas80:


eschnett (approved):
Approved because obviously correct.

(In the end it would be better to use `loop_device` in the source code instead of `loop`.)

rhaas80:
I suspect the issue is that this code won't run on a GPU since it is in `noice.cxx` and creates an instance of `uniform_real_distribution<CCTK_REAL>` which may not exist for GPU code.

PR #371, opened 2025-11-25
Enable subcycle time stepping in CarpetX schedule

lwJi:
Summary Refactored `driver.cxx` and `schedule.cxx` to enable subcycling support in CarpetX.

Change Log

1. Regression Test: Added a WaveToyX test (3 refinement levels) to ensure backward compatibility.

2. Parameter Scope: Promoted `use_subcycling_wip` to RESTRICTED scope for external thorn usage.

3. Driver Logic: Enabled subcycling during regridding (provided levels align) in `driver.cxx`.

4. Schedule Logic: Updated `schedule.cxx` to support the new subcycling flow.

5. Unit Test: Added a new test case specifically for subcycle time stepping.


CarpetX/src/schedule.cxx
```diff
@@ -1023,6 +1025,22 @@ std::vector<clause_t> decode_clauses(const cFunctionData *restrict attribute,
   return result;
 }
 
+CCTK_REAL get_mindx(const bool use_subcycling) {
+  CCTK_REAL mindx = 1.0 / 0.0;
```
rhaas80: might be better to use `std::nan("")` for this (just in case anyone ever tries to run with floating point exceptions enabled).

CarpetX/src/schedule.cxx
```diff
@@ -1023,6 +1025,22 @@ std::vector<clause_t> decode_clauses(const cFunctionData *restrict attribute,
...
+  for (const auto &patchdata : ghext->patchdata) {
+    const amrex::Geometry &geom = patchdata.amrcore->Geom(0);
+    const CCTK_REAL *restrict const dx = geom.CellSize();
+    CCTK_REAL mindx1 = 1.0 / 0.0;
```
rhaas80: same as above (though I suspect `1.0/0.0` was there before, just in a different spot of the file)

CarpetX/src/schedule.cxx
```diff
@@ -1112,6 +1130,14 @@ int Initialise(tFleshConfig *config) {
     CCTK_Traverse(cctkGH, "CCTK_RECOVER_VARIABLES");
     CCTK_Traverse(cctkGH, "CCTK_POST_RECOVER_VARIABLES");
 
+    // Here we assume that all levels have catched up to the coarsed one when
```
rhaas80: "caught" I think. "coarsed" -> "coarse" or "coarsest" (not sure which one is intended, I suspect "coarsest").

CarpetX/src/schedule.cxx
```diff
@@ -1112,6 +1130,14 @@ int Initialise(tFleshConfig *config) {
...
+    // checkpointing.
+    // TODO: checkpoint level.iteration instead.
```
rhaas80, line 1149: is this TODO still relevant? Ie are we happy checkpointing only every coarse step?

CarpetX/src/schedule.cxx
```diff
@@ -1746,21 +1751,14 @@ int Evolve(tFleshConfig *config) {
...
     } // Regrid
 
-    // Find smallest iteration number. Levels at this iteration will
-    // be evolved.
-    rat64 iteration = ghext->patchdata.at(0).leveldata.at(0).iteration;
-    using std::min;
-    for (const auto &patchdata : ghext->patchdata)
-      for (const auto &leveldata : patchdata.leveldata)
-        iteration = min(iteration, leveldata.iteration);
+    cctkGH->cctk_iteration += 1;
```
rhaas80, line 1785: does this work if the finest possible level does not actually exist yet?

CarpetX/src/schedule.cxx
```diff
@@ -1775,18 +1773,31 @@ int Evolve(tFleshConfig *config) {
...
           level_iteration = patchdata.leveldata.at(min_level).iteration;
+          level_delta_iteration =
+              patchdata.leveldata.at(min_level).delta_iteration;
+          break;
+        }
+      }
       assert(level_iteration != -1);
+      assert(level_delta_iteration != -1);
+      // Skip those evolved coarse levels when evolving fine levels to catch up
       if (level_iteration > iteration)
```
rhaas80, line 1821: not sure I understand this comment anymore

CarpetX/src/schedule.cxx
```diff
@@ -1775,18 +1773,31 @@ int Evolve(tFleshConfig *config) {
...
-        break;
+        continue;
 
+      // must not terminate loop iteration due to active_levels being reset at
+      // bootom of loop body
```
rhaas80: "bootom" -> "bottom"

rhaas80: how would we terminate the loop here? How are we not?

CarpetX/src/schedule.cxx
```diff
@@ -1796,6 +1807,11 @@ int Evolve(tFleshConfig *config) {
 
       CycleTimelevels(cctkGH);
 
+      cctkGH->cctk_timefac = use_subcycling_wip ? std::pow(2, min_level) : 1;
+      cctkGH->cctk_time =
+          use_subcycling_wip ? cctkGH->cctk_delta_time * double(level_iteration)
+                             : cctkGH->cctk_time + cctkGH->cctk_delta_time;
```
rhaas80: I wonder if there's a way to avoid accumulating roundoff error here. All possible times are still going to be multiples of the finest possible timestep.

CarpetX/src/schedule.cxx
```diff
@@ -1805,12 +1821,32 @@ int Evolve(tFleshConfig *config) {
       for (int level = ghext->num_levels() - 2; level >= 0; --level)
         Reflux(cctkGH, level);
 
+      // reset active_levels to all that have caught to the same time
```
rhaas80: "to" -> "of"?

CarpetX/src/schedule.cxx
```diff
@@ -1805,12 +1821,32 @@ int Evolve(tFleshConfig *config) {
...
-        for (int level = ghext->num_levels() - 2; level >= 0; --level)
-          Restrict(cctkGH, level);
-        CCTK_Traverse(cctkGH, "CCTK_POSTRESTRICT");
+        active_levels->loop_fine_to_coarse([&](const auto &leveldata) {
+          if (leveldata.level + 1 < active_levels->max_level) {
+            Restrict(cctkGH, leveldata.level);
+            restricted = true;
+          }
+        });
+        // Perform actions if the current level was restricted
```
rhaas80: perform what action? The comment basically says "Do something" which is not helpful

CarpetX/src/schedule.cxx
```diff
@@ -1822,7 +1858,7 @@ int Evolve(tFleshConfig *config) {
       total_evolution_output_time += output_finish_time - output_start_time;
 
       active_levels = optional<active_levels_t>();
```
rhaas80: this is the one we need to encounter and why a "continue" must not be used? I wonder if this could be incorporated into the loop header itself.

TestSubcycling/doc/documentation.tex
```diff
@@ -0,0 +1,146 @@
...
+
+
+\begin{thebibliography}{9}
+
+\end{thebibliography}
+
+% Do not delete next line
+% END CACTUS THORNGUIDE
+
+\end{document}
```
rhaas80: this is basically empty. I'd at least add an abstract or alternatively remove the file.

TestSubcycleStepping/par/subcycle.par
```diff
@@ -0,0 +1,54 @@
...
+  TestSubcycling
+"
+
+Cactus::presync_mode = "mixed-error"
+
+CarpetX::poison_undefined_values = yes
+
+CarpetX::ncells_x = 32
+CarpetX::ncells_y = 32
+CarpetX::ncells_z = 32
```
rhaas80, line 14: could we get a away with a smaller grid (and thus smaller output files)?

rhaas80: https://docs.einsteintoolkit.org/et-docs/Adding_a_test_case

TestSubcycleStepping/par/subcycle.par
```diff
@@ -0,0 +1,54 @@
...
+CarpetX::max_tile_size_x = 10000000
+CarpetX::max_tile_size_y = 10000000
+CarpetX::max_tile_size_z = 10000000
+
+CarpetX::dtfac = 16.0
+
+Cactus::cctk_itlast = 8
+
+CarpetX::use_subcycling_wip = yes
+CarpetX::restrict_during_sync = no
```
rhaas80, line 33: add an assert that `use_subcycling_wip` and `restrict_during_sync` cannot be true at the same time (could they both be false?)?

TestSubcycleStepping/test/subcycle/carpetxregrid-regrid_error.it000000.x.tsv
rhaas80, line 1: we could probably keep only the `.y.` files, at the cost of possibly missing bugs that affect only restriction / prolongation in the x or z directions.

TestSubcycling/test/subcycle/testsubcycling-iteration.it000008.z.tsv
rhaas80, line 1: do we need the iterations other than 0 and the last one for the regression test?

TestSubcycling/README
```diff
@@ -0,0 +1,9 @@
+Cactus Code Thorn TestSubcycling
+Author(s)    : Roland Haas <rhaas@illinois.edu>
+Maintainer(s): Roland Haas <rhaas@illinois.edu>
+Licence      : NCSA
+--------------------------------------------------------------------------
+
+1. Purpose
+
+not documented
```
rhaas80: Something should be documented as a purpose

WaveToyX/test/radiative-3levels.par
rhaas80, line 1: do we need all these files, or would eg the ".y." ones be sufficient? Along with only some of the norms?

The test code could still output more, we just would not store them and include them in the set of files to compare against.

WaveToyX/test/radiative-3levels.par
```diff
@@ -0,0 +1,55 @@
...
+
+WaveToyX::initial_condition = "Gaussian"
+WaveToyX::gaussian_width = 0.5
+WaveToyX::boundary_condition = "radiative"
+
+ODESolvers::method = "RK3"
+
+IO::out_dir = $parfile
+IO::out_every = 10
+IO::parfile_write = no
```
rhaas80, line 43: should add:

```
 IO::out_fileinfo  = "axis labels"
```

even though it may not actually be used by any current CarpetX output routine.

rhaas80 (changes requested):

CarpetX/src/schedule.cxx
lwJi: For now, we are happy to only checkpointing every coarse step. But I still want keep this TODO, because it will enable us to checkpoint at any time step

CarpetX/src/schedule.cxx
lwJi: it should be "to"

rhaas80: so ` // reset active_levels to all those that have caught to up the same time` ? We need a native English speaker here :-)

CarpetX/src/schedule.cxx
yosef-zlochower: note mindx=fmin(mindx, blah). So setting mindx to inf should do something sensible. Will fmin(nan, blah) work?

rhaas80: yes, this works fine and is standards conforming. `min` and `max` are special for NaN (in that they are not "infected" by NaN in their inputs).

https://en.cppreference.com/w/c/numeric/math/fmin says "Returns the smaller of two floating-point arguments, treating NaNs as missing data (between a NaN and a numeric value, the numeric value is chosen)."

CarpetX/src/schedule.cxx
lwJi: This `continue` effectively select `active_levels` which are evolving during the current `cctk_iteration`.

Suppose we have 3 levels, and current doing `cctk_iteration=3` iteration. Let's look at this `min_level, max_level` loop:

1. `min_level=0, max_level=1`: at this point, `level_iteration=4/4` while `iteration=2/4`, so `continue` is triggered. We are going to the next `min_level, max_level` iteration.
2. `min_level=1, max_level=2`: at this point, `level_iteration=2/4` while `iteration=2/4`. so `continue` isn't triggered. it calls `CCTK_EVOL` on `active_levels={1}`.
3. `min_level=2, max_level=3`: at this point, `level_iteration=2/4` while `itertaion=2/4`. so `continue` isn't triggerd. it calls `CCTK_EVOL` on `active_levels={2}`.
4. `min_level=3, max_level=4`: it doesn't agree with `min_level < ghext->num_levels()` anymore, so the loop stops.

we finish the `cctk_iteration=3` and going to the next `cctk_iteration`.

CarpetX/src/schedule.cxx
lwJi: fixed

CarpetX/src/schedule.cxx
lwJi: fixed

CarpetX/src/schedule.cxx
lwJi: fixed

CarpetX/src/schedule.cxx
lwJi: Please take a look at the fixed version. I think it's close to 'native' Enghlish speaker ;)

CarpetX/src/schedule.cxx
lwJi: Please take a look at the above 3 levels example

CarpetX/src/schedule.cxx
lwJi: There are two iteration system here (3 levels):
1. cctk_iteration:
```
----- ----- -----
              4
        3   -----
              3
  1   ----- -----
              2
        1   -----
              1
----- ----- -----
```
2. level_iteration:
```
----- ----- -----
             4/4
       4/4  -----
             3/4
 4/4  ----- -----
             2/4
       2/4  -----
             1/4
----- ----- -----
```

CarpetX/src/schedule.cxx
lwJi: `active_levels` is set right before it was used above. Here we reset it to empty to prevent accident use somehow.

CarpetX/src/schedule.cxx
lwJi: I'm all ears

CarpetX/src/schedule.cxx
lwJi: I think that depends if it make sense when we pass the nonexist finest to constructor of `active_levels`

TestSubcycling/doc/documentation.tex
lwJi: removed

TestSubcycling/README
lwJi: done

TestSubcycling/test/subcycle/testsubcycling-iteration.it000008.z.tsv
lwJi: Since there are 2 levels in the test, a complete cycle (every level catch up a coarest iteration) is 2 iterations. But I think it's a good idea to keep at least 2 cycles, which is 4 iterations.

So I'd like to keep iteration {0, 1, 2, 3, 4}.

We should keep iteration {1, 3} too, because we care about what the driver take us during 'substeps'.

TestSubcycleStepping/par/subcycle.par
lwJi: very good. I have set ncells = 16x16x16.
I can't make it 8x8x8, because it will refine most of the regions, other than just the interior.

TestSubcycleStepping/test/subcycle/carpetxregrid-regrid_error.it000000.x.tsv
lwJi: Come on, these files are pretty small. Let's keep all x,y,z

TestSubcycleStepping/par/subcycle.par
lwJi: I add an assert at Initialize to make sure `restrict_during_sync=no` when `use_subcycling_wip=yes`.

`restrict_during_sync` was there before we turn on subcycling. I think restriction is a option when you do mesh refinement. You can in principle choose not to restrict.

WaveToyX/test/radiative-3levels.par
lwJi: I have add this par. Could you explain what does this par do?

WaveToyX/test/radiative-3levels.par
lwJi: I think you argument applies to other WaveToyX tests too. I would modify them together but it's outside the scope of this PR

CarpetX/param.ccl
```diff
@@ -95,6 +95,13 @@ BOOLEAN periodic_z "Periodic boundary conditions in z-direction"
   "no"  :: "not periodic in z direction"
 } "yes"
 
+RESTRICTED:
+
+BOOLEAN use_subcycling_wip "Use subcycling in time"
```
eschnett: It is usually not a good idea to let other thorns look at parameters. Parameters are how people communicate with a thorn, and the meaning of parameters can change over time when we introduce a better "user interface".

I assume that other thorns want to know whether CarpetX is using subcycling. It is best to introduce a function to query this, e.g. near the `num_patches()` function in `driver.hxx`.

CarpetX/src/schedule.cxx
eschnett: This is infinity, not nan, and should not trigger floating point exceptions. You can use `std::numeric_limits<CCTK_REAL>::infinity()` instead.

CarpetX/src/schedule.cxx
```diff
@@ -1112,6 +1135,14 @@ int Initialise(tFleshConfig *config) {
     CCTK_Traverse(cctkGH, "CCTK_RECOVER_VARIABLES");
     CCTK_Traverse(cctkGH, "CCTK_POST_RECOVER_VARIABLES");
 
+    // Here we assume that all levels have caught up to the coarsest one when
+    // checkpointing.
+    // TODO: checkpoint level.iteration instead.
+    const int iteration_ratio = pow(2, ghext->num_levels() - 1);
```
eschnett: This calls `std::pow`, which returns a `double`, and which is not guaranteed to be exact. The conversion to `int` might generate a wrong result. Either round to nearest, or use `Arith::pown` (which returns `int`), or use the operator `<<`.

CarpetX/src/schedule.cxx
```diff
@@ -1122,17 +1153,7 @@ int Initialise(tFleshConfig *config) {
...
-        const amrex::Geometry &geom = patchdata.amrcore->Geom(0);
-        const CCTK_REAL *restrict const dx = geom.CellSize();
-        CCTK_REAL mindx1 = 1.0 / 0.0;
-        for (int d = 0; d < dim; ++d)
-          mindx1 = fmin(mindx1, dx[d]);
-        mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
-        mindx = fmin(mindx, mindx1);
-      }
-      cctkGH->cctk_delta_time = dtfac * mindx;
+      cctkGH->cctk_delta_time = dtfac * get_mindx(use_subcycling_wip);
```
eschnett: Whether subcycling is enabled or not is a global setting that doesn't change over time. There is no need to pass this as argument to `get_mindx` which already looks at other settings and other global state.

What does `mindx` mean in this context? Is this still the right name for the function? Is this the minimum spacing of the coarsest level, the current level, the finest level, the finest currently active level, ...?

Maybe it's cleanest to write `use_subcycling ? get_coarse_mindx() : get_finest_mindx()` here.

CarpetX/src/schedule.cxx
```diff
@@ -1367,6 +1368,8 @@ int Initialise(tFleshConfig *config) {
       if (leveldata.level != ghext->num_levels() - 1)
         Restrict(cctkGH, leveldata.level);
     });
+    // Prolongation
+    SyncAfterRestrict(cctkGH);
```
eschnett: This sounds weird. What does `SyncAfterRestrict` do? The "after restrict" sounds as if this function is supposed to be called after restricting. That is a quite uncommon way to name a function. Functions are usually named after what they do. Since restricting changes the coarser levels, I assume you want to sync and prolongate to the finer levels. The comment "Prolongate" also indicates that the function prolongates. Why not just call it `SyncAndProlongate`?

CarpetX/src/schedule.cxx
```diff
@@ -1638,6 +1647,17 @@ int Evolve(tFleshConfig *config) {
...
 
       for (const auto &patchdata : ghext->patchdata) {
+
+        int min_active_level = -1;
+        for (int min_level = patchdata.leveldata.size() - 1; min_level >= 0;
+             --min_level)
+          if (patchdata.leveldata.at(min_level).iteration == iteration) {
+            min_active_level = min_level;
+          } else {
+            break;
```
eschnett, line 1672: That's a complicated logic. Invert the `if` condition, leave only the `break` in the `if`, and have the assignment after the `if` statement.

CarpetX/src/schedule.cxx
```diff
@@ -1638,6 +1647,17 @@ int Evolve(tFleshConfig *config) {
       Interval interval(timer);
 
       for (const auto &patchdata : ghext->patchdata) {
+
+        int min_active_level = -1;
+        for (int min_level = patchdata.leveldata.size() - 1; min_level >= 0;
+             --min_level)
```
eschnett: Please add braces around multi-line `for` bodies.

CarpetX/src/schedule.cxx
```diff
@@ -1638,6 +1647,17 @@ int Evolve(tFleshConfig *config) {
       Interval interval(timer);
 
       for (const auto &patchdata : ghext->patchdata) {
+
+        int min_active_level = -1;
+        for (int min_level = patchdata.leveldata.size() - 1; min_level >= 0;
```
eschnett: This just loops over levels. Please rename the iteration variable to `level`.

CarpetX/src/schedule.cxx
eschnett: I introduced the concepts of "batches of levels" when iterating. Do you still support this? If so, did you test it? If not, please remove it -- we can reintroduce it later. There is no point in keeping complicated logic around if it is broken.

CarpetX/src/schedule.cxx
```diff
@@ -1775,18 +1778,31 @@ int Evolve(tFleshConfig *config) {
...
+      }
       assert(level_iteration != -1);
+      assert(level_delta_iteration != -1);
+      // Skip those evolved coarse levels when evolving fine levels to catch up
       if (level_iteration > iteration)
-        break;
+        continue;
 
+      // must not terminate loop iteration due to active_levels being reset at
+      // bottom of loop body
```
eschnett: What does this mean? Why would resetting an optional variable terminate a loop iteration? And if you do it at the bottom of the loop body, doesn't the loop iteration "terminate" there any way? The expression "terminate a loop iteration" is funny: usually one would speak of either "breaking out of a loop" (or "exiting a loop"), or maybe "terminating a loop", but skipping the remainder of the loop body -- what `continue` does -- is usually not called "terminating".

And what you are doing here is allocating an optional value. How is this related to resetting it?

CarpetX/src/schedule.cxx
eschnett: We're using rational number here. The current time is always the iteration number times the time step size.

CarpetX/src/schedule.cxx
```diff
@@ -1805,12 +1826,33 @@ int Evolve(tFleshConfig *config) {
...
-        for (int level = ghext->num_levels() - 2; level >= 0; --level)
-          Restrict(cctkGH, level);
-        CCTK_Traverse(cctkGH, "CCTK_POSTRESTRICT");
+        active_levels->loop_fine_to_coarse([&](const auto &leveldata) {
+          if (leveldata.level + 1 < active_levels->max_level) {
+            Restrict(cctkGH, leveldata.level);
+            restricted = true;
+          }
+        });
+        // Perform CCTK_POSTRESTRICT if the current level was restricted
```
eschnett: I don't think this checks the "current level", whatever that even may be. I think this checks whether any level was restricted to. The comment is wrong.

CarpetX/src/schedule.cxx
```diff
@@ -1805,12 +1826,33 @@ int Evolve(tFleshConfig *config) {
...
+
       if (!restrict_during_sync) {
+        bool restricted = false;
         // Restrict
-        // TODO: These loop bounds are wrong for subcycling
-        for (int level = ghext->num_levels() - 2; level >= 0; --level)
-          Restrict(cctkGH, level);
-        CCTK_Traverse(cctkGH, "CCTK_POSTRESTRICT");
+        active_levels->loop_fine_to_coarse([&](const auto &leveldata) {
+          if (leveldata.level + 1 < active_levels->max_level) {
```
eschnett: What condition are you checking here? Are you assuming that there can be inactive levels below the set of active levels, i.e. that there are finer-but-inactive levels? I don't think that can be true, that would not make sense. At each iteration a certain set of levels is active, in the sense that levels which are too coarse and which don't exist at the current time are inactive.

The `Restrict` function already checks whether the source level exists. I think this if statement is superfluous, and the `restricted` flag is not necessary.

CarpetX/src/schedule.cxx
rhaas80: still supported. For sure the extreme cases (all levels in one batch, and one level per batch) are tested. The middle-ground cases may or may not be tested (depends on whether someone set up a run with non-default time-refinement-factors).

rhaas80: > This `continue` effectively select `active_levels` which are evolving during the current `cctk_iteration`.
> 
> Suppose we have 3 levels, and current doing `cctk_iteration=3` iteration. Let's look at this `min_level, max_level` loop:
> 
>     1. `min_level=0, max_level=1`: at this point, `level_iteration=4/4` while `iteration=2/4`, so `continue` is triggered. We are going to the next `min_level, max_level` iteration.
> 
>     2. `min_level=1, max_level=2`: at this point, `level_iteration=2/4` while `iteration=2/4`. so `continue` isn't triggered. it calls `CCTK_EVOL` on `active_levels={1}`.
> 
>     3. `min_level=2, max_level=3`: at this point, `level_iteration=2/4` while `itertaion=2/4`. so `continue` isn't triggerd. it calls `CCTK_EVOL` on `active_levels={2}`.
> 
>     4. `min_level=3, max_level=4`: it doesn't agree with `min_level < ghext->num_levels()` anymore, so the loop stops.
> 
> 
> we finish the `cctk_iteration=3` and going to the next `cctk_iteration`.

It's not the code that I don't understand. It's the comment :-)

CarpetX/src/schedule.cxx
rhaas80: `cctk_time` and `cctk_delta_time` are doubles though, so the expression (the non-subcycling one) `                             : cctkGH->cctk_time + cctkGH->cctk_delta_time` will accumulate roundoff error. 

In fact, shouldn't `cctkGH->cctk_delta_time * double(level_iteration)` be correct for all cases (except for adaptive time-stepping in which case all bets are off anyway)?

WaveToyX/test/radiative-3levels.par
rhaas80: it causes only the variable names to be output in the header of the ASCII file output, and not eg build id and run time (which would differ each time the simulations is run or the code compiled). Making for tidier commits.

CarpetX/src/schedule.cxx
```diff
@@ -2848,6 +2890,24 @@ void Restrict(const cGH *cctkGH, int level) {
...
 }
 
+void SyncAfterRestrict(const cGH *cctkGH) {
+  const int numgroups = CCTK_NumGroups();
+  vector<int> groups;
+  groups.reserve(numgroups);
+  const auto &patchdata0 = ghext->patchdata.at(0);
+  const auto &leveldata0 = patchdata0.leveldata.at(0);
+  for (const auto &groupdataptr : leveldata0.groupdata) {
+    // Sync only grid functions
```
eschnett, line 2924: We have accumulated logic in the current synchronization/prolongation functions that ensure that all grid functions remain consistent, and we track which parts of what grid function is valid.

Here (that is, in the function `Restrict` above) you make a decision which grid functions to restrict. This means that the other grid functions are either constant in time (coordinates) or are now invalid (since they have not been restricted). You need to track this. I think you need to talk to the time integrator for this; you should restrict all those grid functions which are integrated in time, and mark all of their dependencies as invalid.

Also, since you modified these grid functions, they will have to have their boundary conditions applied again. It is probably easiest to schedule `ODESolvers_PostStep` there (this schedule groups is misnamed) because this ensures that everything is set up correctly and recalculated as necessary.

That also means that `SyncAfterRestrict` will not be needed any more.

WaveToyX/test/radiative-3levels.par
```diff
@@ -0,0 +1,55 @@
...
+    ODESolvers
+    WaveToyX
+"
+
+Cactus::cctk_show_schedule = no
+Cactus::presync_mode = "mixed-error"
+
+CarpetX::poison_undefined_values = yes
+
+CarpetX::ncells_x = 8
```
eschnett: A refinement test with just 8 cells? Is there actually any refinement happening? If so, are you sure that any coarse cells remain that are not covered by refined cells?

CarpetX/src/schedule.cxx
rhaas80: I misplaced my own comment by one line.

If I recall correctly the comment wants to say that past this point you *must* reach the end of the loop (and the reset of `active_levels` there) and must not use `continue` or `break` to short-circuit a loop. 

Better wording is required for the comment.

CarpetX/src/schedule.cxx
rhaas80: "current batch of levels"

TestSubcycling/schedule.ccl
```diff
@@ -0,0 +1,16 @@
+# Schedule definitions for thorn TestSubcycling
+
+STORAGE: iteration
+
+schedule TestSubcycling_Init AT INITIAL
```
eschnett: This thorn tests subcycling without using ODESolvers. I think it is important to also  have a test that uses ODESolvers because this is what most people will do.

I am also surprised that this thorn does not schedule anything in POSTREGRID or POSTRESTRICT. If this thorn performs basic tests then it should test these bins as well. Wouldn't it at least need to apply boundary conditions in these schedule bins?

CarpetX/src/schedule.cxx
rhaas80: Just in case "Sync only grid functions" here means "don't sync grid arrays and grid scalars" (which are never synced).

CarpetX/param.ccl
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done. you are absolutely right

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: done

CarpetX/src/schedule.cxx
lwJi: This condition is trying to skip the finest level, because there is no finer level it can restrict from. But since `Restrict` already check whether the source level exists, this if statement is superfluous for calling `Restrict`.

However, the flag is necessary, because we don't want call `CCTK_POSTRESTRICT` on the finest level.
I will move `Restrict()` outside the if statement.

WaveToyX/test/radiative-3levels.par
lwJi: I see. thanks

CarpetX/src/schedule.cxx
lwJi: Suppose we have two levels and we are at cctk_iteration 1, there are two steps happening during this cctk_iteration
1. the coarse level step first, at this point, the fine level is still at cctk_iteration 0. if we `Restrict`, it would pull the data from wrong time step.
2. the fine level step next, as you said, since there is no source, `Restrcit` will do nothing.

So the if statement is trying to prevent pulling wrong data at step 1.

CarpetX/src/schedule.cxx
lwJi: by using `if (groupdata.do_checkpoint)`, we are only `SyncAndProlongate` those grid functions which are integrated in time. Marking all of their dependencies as invalid is done inside `SyncGroupsByDirI`. And applying boundary conditions is also done inside `SyncGroupsByDirI`.

WaveToyX/test/radiative-3levels.par
lwJi: you are absolutely right. Since the default blocking_factor=8. it had only 2 levels effectively. Now I modify ncells=16, it has well nested 3 levels (I have checked the data ;)).
And I regenerated the test data used `main` branch, so this test makes sure this PR will not modify the non-subcycling behavior.

TestSubcycling/schedule.ccl
lwJi: The intention of this test is only to test each level is moving forward correctly in subcycling case. There is no true time integration implemented yet (RK, Euler, etc.). In the next PR, I will add the time integration part and add the corresponding ODESolver tests (where apply boundary conditions are applied through SYNC).

CarpetX/src/schedule.cxx
lwJi: @rhaas80 I need you help with this comment ;)

lwJi: We removed this comment because it created more confusion than clarity.

CarpetX/src/schedule.cxx
rhaas80: @eschnett ^^ ?

CarpetX/src/schedule.cxx
rhaas80: @eschnett : poke

TestSubcycling/schedule.ccl
rhaas80: Do we have a test that tests the full set of functionality? Eg a wavetoy with subcycling?

CarpetX/src/schedule.cxx
```diff
@@ -1023,6 +1025,33 @@ std::vector<clause_t> decode_clauses(const cFunctionData *restrict attribute,
...
+
+CCTK_REAL get_finest_mindx() {
+  CCTK_REAL mindx = std::numeric_limits<CCTK_REAL>::infinity();
+  for (const auto &patchdata : ghext->patchdata) {
+    const amrex::Geometry &geom = patchdata.amrcore->Geom(0);
+    const CCTK_REAL *restrict const dx = geom.CellSize();
+    CCTK_REAL mindx1 = std::numeric_limits<CCTK_REAL>::infinity();
+    for (int d = 0; d < dim; ++d)
+      mindx1 = fmin(mindx1, dx[d]);
+    mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
```
rhaas80, line 1047: `ldexp` seems to be a somewhat obscure function. Not sure of the benefit of efficiency here is worth the non-intuitive name. This was a `pow(2, -n)` before I guess? Since CarpetX already requires Arith, why not use its `pown`?

TestSubcycling/schedule.ccl
lwJi: We do have one. I can add it if you don't mind this pr contains too many files ;)

CarpetX/src/schedule.cxx
```diff
@@ -1122,17 +1164,9 @@ int Initialise(tFleshConfig *config) {
...
-        CCTK_REAL mindx1 = 1.0 / 0.0;
-        for (int d = 0; d < dim; ++d)
-          mindx1 = fmin(mindx1, dx[d]);
-        mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
-        mindx = fmin(mindx, mindx1);
-      }
-      cctkGH->cctk_delta_time = dtfac * mindx;
+      cctkGH->cctk_delta_time =
+          dtfac *
+          (ghext->use_subcycling() ? get_coarse_mindx() : get_finest_mindx());
```
rhaas80: without subcycling, wouldn't `get_coarse_mindx` be the same as `get_finest_mindx`?

CarpetX/src/schedule.cxx
```diff
@@ -1122,17 +1164,9 @@ int Initialise(tFleshConfig *config) {
...
     } else {
       abort();
```
rhaas80, line 1170: not your code but: better error reporting here would be nice. Or at least a comment that this is a bug in handling all possible cases of the keyword.

CarpetX/src/schedule.cxx
```diff
@@ -1367,6 +1385,8 @@ int Initialise(tFleshConfig *config) {
       if (leveldata.level != ghext->num_levels() - 1)
         Restrict(cctkGH, leveldata.level);
     });
+    // Prolongation
+    SyncAndProlongate(cctkGH);
```
rhaas80: Is the use of "Sync" == "Ghost Zone Exchange" consistent with the remainder of CarpetX? Or Does this need to be `FullSync` or `ExchangeGhostsAndApplyBCAndProlongate`???

CarpetX/src/schedule.cxx
rhaas80: ok, it seems we would definitely benefit for a test case that exercises some of the corner cases of the algorithm. Does not have to be in this pull request.

TestSubcycling/schedule.ccl
lwJi: This pr only add stepping functionality which has been tested in TestSubycling. in the coming two PRs
1. add subcycling algorithm support, where I will add a 'wavetoy' example to test the "full set of functionality"
2. add subcycling algorithm support inside ODESolvers, where I will add another 'wavetoy' example with ODESolvers to test the "full set of functionality"

CarpetX/src/schedule.cxx
eschnett: Let's assume there are three levels. The finest level is never restricted to, so nothing happens at the quarter iterations.

At iteration 0, levels 0 and 1 are restricted to.
At iteration 1/2, only level 1 is restricted to.
At iteration 1, again levels 0 and 1 are restricted to.

I think you are splitting what happens e.g. at iteration 0 or 1 into two parts. That's not necessary since you start restricting from the finest level.

At each iteration you wait until all levels that are active at that iteration have stepped ahead in time, and then they all get restricted.

CarpetX/src/schedule.cxx
eschnett: This looks good.

TestSubcycling/schedule.ccl
eschnett: Fair enough.

Still, you need to apply boundary conditions (i.e. synchronize) at a few other points:
- CarpetX will regrid while setting up initial conditions on multiple levels. You need to sync after regridding.
- Are you testing regridding during evolution? Should this work?
- Are you supporting checkpointing/recovery? Should this work?

You are setting up initial conditions on all grid points, but you are evolving only the interior. This creates a difference because the boundary conditions are not applied to the initial conditions. It would be better to synchronize after setting up initial conditions instead of setting up initial conditions everywhere.

CarpetX/src/schedule.cxx
```diff
@@ -1821,8 +1881,10 @@ int Evolve(tFleshConfig *config) {
       const double output_finish_time = gettime();
       total_evolution_output_time += output_finish_time - output_start_time;
 
-      active_levels = optional<active_levels_t>();
-    } // for min_level
+    } // for min_level, max_level
+
+    // Mark all levels inactive
```
rhaas80: this comment explains what the code does (which one can get from reading), but not why (which one cannot). Something like "Mark all level inactive now that we are done processing a time step" may be more helpful. The problem is the the global `active_levels` variable is part of the input to various function but we cannot actually pass it as an argument to them.

TestSubcycleStepping/test/subcycle.par
```diff
@@ -0,0 +1,56 @@
...
+CarpetX::out_silo_every = 0
+CarpetX::out_openpmd_every = 0
+
+CarpetX::verbose = no
+CarpetX::out_performance = no
+CarpetX::out_metadata = no
+CarpetX::out_tsv_vars = "
+  TestSubcycling::iteration
+  CarpetXRegrid::regrid_error
+"
```
rhaas80, line 56: do we need `.x.`, `.y.` and `.z.` output for all variables? Or would `.y.` be sufficient alone to guard against regressions?

rhaas80 (commented):
no further objections from me. But I have not looked deeply into the parts that Erik already commented on.

CarpetX/src/schedule.cxx
lwJi: @eschnett , as suggested, we only restrict when all active levels up to finest one catch up each other, and get rid of the old if statement

TestSubcycling/schedule.ccl
lwJi: @eschnett 

> * CarpetX will regrid while setting up initial conditions on multiple levels. You need to sync after regridding.

as suggested, we sync to apply boundary conditions when setting initial data.

> * Are you testing regridding during evolution? Should this work?
> * Are you supporting checkpointing/recovery? Should this work?

not in this test. but they both work in our BBH tests

CarpetX/src/schedule.cxx
lwJi: this was there before in the `main` branch. I would claim it's beyond the scope of this PR. hahaha

CarpetX/src/schedule.cxx
lwJi: no. `get_coarse_mindx` would always get the dx of coarsest level, while `get_finest_mindx` would get the dx of the finest level

CarpetX/src/schedule.cxx
lwJi: I have added `CCTK_ERROR("Unsupported value for 'timestep_choice'");` before abort.

CarpetX/src/schedule.cxx
lwJi: I have rename it to `FullSync`

CarpetX/src/schedule.cxx
lwJi: done

TestSubcycleStepping/test/subcycle.par
lwJi: I would still claim that the files are pretty small ;)

CarpetX/src/schedule.cxx
```diff
@@ -83,6 +83,7 @@ std::optional<active_levels_t> active_levels;
 void Reflux(const cGH *cctkGH, int level);
 void Restrict(const cGH *cctkGH, int level, const std::vector<int> &groups);
 void Restrict(const cGH *cctkGH, int level);
+void FullSync(const cGH *cctkGH);
```
rhaas80: name is not fully intuitive to me

CarpetX/src/schedule.cxx
```diff
@@ -1023,6 +1025,33 @@ std::vector<clause_t> decode_clauses(const cFunctionData *restrict attribute,
...
 
+CCTK_REAL get_coarse_mindx() {
+  CCTK_REAL mindx = std::numeric_limits<CCTK_REAL>::infinity();
+  for (const auto &patchdata : ghext->patchdata) {
+    const amrex::Geometry &geom = patchdata.amrcore->Geom(0);
+    const CCTK_REAL *restrict const dx = geom.CellSize();
+    CCTK_REAL mindx1 = std::numeric_limits<CCTK_REAL>::infinity();
+    for (int d = 0; d < dim; ++d)
+      mindx1 = fmin(mindx1, dx[d]);
+    mindx = fmin(mindx, mindx1);
```
rhaas80: why have the `mindx` and `mindx1` split? I could see that for multi-threading but it does not seem required here, is it?

CarpetX/src/driver.hxx
```diff
@@ -450,6 +450,10 @@ struct GHExt {
   };
   std::vector<PatchData> patchdata; // [patch]
 
+  bool do_subcycling = false; // whether to subcycling
```
eschnett: I recommend a single field `use_subcycling` and no function to read it. This would be a consistent naming scheme.

With this patch, is subcycling working? If so, remove the `_wip` (work-in-progress) suffix from the parameter name.

CarpetX/src/schedule.cxx
```diff
@@ -1039,6 +1068,11 @@ int Initialise(tFleshConfig *config) {
     CCTK_ERROR("CarpetX currently requires Cactus::presync_mode = "
                "\"mixed-error\" or \"presync-only\"");
 
+  // Check restrict_during_sync when use_subcycling is on
+  if (ghext->use_subcycling() && restrict_during_sync)
+    CCTK_ERROR("CarpetX currently requires CarpetX::restrict_during_sync = "
```
eschnett: Remove "currently"?

CarpetX/src/schedule.cxx
```diff
@@ -1160,18 +1195,11 @@ int Initialise(tFleshConfig *config) {
...
-            mindx1 = fmin(mindx1, dx[d]);
-          mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
-          mindx = fmin(mindx, mindx1);
-        }
-        cctkGH->cctk_delta_time = dtfac * mindx;
+        cctkGH->cctk_delta_time =
+            dtfac *
+            (ghext->use_subcycling() ? get_coarse_mindx() : get_finest_mindx());
       } else {
+        CCTK_ERROR("Unsupported value for 'timestep_choice'");
```
eschnett: If this is a keyword parameter then this would be an "internal error" here because the if statement is supposed to handle all possible values. Otherwise, if this error can actually happen, then the value should be output to make life simpler for those encountering the error. Probably call it `CarpetX::timestep_choice` then, or wherever the parameter comes from.

CarpetX/src/schedule.cxx
```diff
@@ -1321,18 +1349,11 @@ int Initialise(tFleshConfig *config) {
...
-                mindx1 = fmin(mindx1, dx[d]);
-              mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
-              mindx = fmin(mindx, mindx1);
-            }
-            cctkGH->cctk_delta_time = dtfac * mindx;
+            cctkGH->cctk_delta_time =
+                dtfac * (ghext->use_subcycling() ? get_coarse_mindx()
+                                                 : get_finest_mindx());
           } else {
+            CCTK_ERROR("Unsupported value for 'timestep_choice'");
```
eschnett: Ditto.

CarpetX/src/schedule.cxx
```diff
@@ -1367,6 +1388,8 @@ int Initialise(tFleshConfig *config) {
       if (leveldata.level != ghext->num_levels() - 1)
         Restrict(cctkGH, leveldata.level);
     });
+    // Prolongation
+    FullSync(cctkGH);
```
eschnett: Remind me: what does `FullSync` do, compared to a regular `Sync`?

CarpetX/src/schedule.cxx
```diff
@@ -1775,18 +1806,29 @@ int Evolve(tFleshConfig *config) {
...
+        if (min_level < int(patchdata.leveldata.size())) {
           level_iteration = patchdata.leveldata.at(min_level).iteration;
+          level_delta_iteration =
+              patchdata.leveldata.at(min_level).delta_iteration;
+          break;
+        }
+      }
       assert(level_iteration != -1);
+      assert(level_delta_iteration != -1);
+      // Skip those evolved coarse levels when evolving fine levels to catch up
```
eschnett: Wrong grammar.

CarpetX/src/schedule.cxx
```diff
@@ -1796,6 +1838,12 @@ int Evolve(tFleshConfig *config) {
 
       CycleTimelevels(cctkGH);
 
+      cctkGH->cctk_timefac = ghext->use_subcycling() ? (1 << min_level) : 1;
```
eschnett: Is the time refinement factor always 2^level?

CarpetX/src/schedule.cxx
```diff
@@ -2848,6 +2913,24 @@ void Restrict(const cGH *cctkGH, int level) {
   Restrict(cctkGH, level, groups);
 }
 
+void FullSync(const cGH *cctkGH) {
```
eschnett: Does this function synchronize all variables? Is that what "full" means? If so, maybe rename to `SyncAllGFs`?

eschnett (approved):
I made a few suggestions. Please look at them, then merge.

CarpetX/src/schedule.cxx
```diff
@@ -1160,18 +1195,11 @@ int Initialise(tFleshConfig *config) {
...
-          CCTK_REAL mindx1 = 1.0 / 0.0;
-          for (int d = 0; d < dim; ++d)
-            mindx1 = fmin(mindx1, dx[d]);
-          mindx1 = ldexp(mindx1, -(int(patchdata.leveldata.size()) - 1));
-          mindx = fmin(mindx, mindx1);
-        }
-        cctkGH->cctk_delta_time = dtfac * mindx;
+        cctkGH->cctk_delta_time =
+            dtfac *
+            (ghext->use_subcycling() ? get_coarse_mindx() : get_finest_mindx());
```
rhaas80: it's been a while. Do we set `cctkGH->cctk_timefac` and `cctkGH->cctk_levfac` (probably needs to happen in `enter/leave_level_mode` or so.

CarpetX/src/schedule.cxx
rhaas80: should then say "Unexpected" rather than "Unsupported" (since the latter would imply that the option value is actually known, just not supported).

CarpetX/src/schedule.cxx
```diff
@@ -1638,6 +1667,16 @@ int Evolve(tFleshConfig *config) {
...
       for (const auto &patchdata : ghext->patchdata) {
+
+        int min_active_level = -1;
+        for (int level = patchdata.leveldata.size() - 1; level >= 0; --level) {
+          if (patchdata.leveldata.at(level).iteration != iteration) {
+            break;
+          }
+          min_active_level = level;
+        }
+        assert(min_active_level != -1);
```
rhaas80, line 1676: Erik had pointed out (correctly out one assumes) that once one has found the coarsest level that steps forward this iteration then all levels finer than that one must also step forward, which could be added as an `assert()` on the logic of the code here.

rhaas80: though, looking at the `// Find end of batch` comment below, those that have a different timestep size would not be part of the batch, but of a different batch that will step this iteration.

CarpetX/src/schedule.cxx
```diff
@@ -1746,21 +1778,20 @@ int Evolve(tFleshConfig *config) {
...
-        iteration = min(iteration, leveldata.iteration);
+        delta_iteration = min(delta_iteration, leveldata.delta_iteration);
+
+    cctkGH->cctk_iteration += 1;
 
     // Loop over all levels, in batches that combine levels that don't
     // subcycle. The level range is [min_level, max_level).
-    int min_level = 0;
-    while (min_level < ghext->num_levels()) {
+    for (int min_level = 0, max_level = min_level + 1;
```
rhaas80: looking at this, this is a bit confusing, since `max_level` is also updated inside of the loop body. It cannot live only inside of the loop since we need `max_level` to update `min_level`.

CarpetX/src/schedule.cxx
```diff
@@ -1746,21 +1778,20 @@ int Evolve(tFleshConfig *config) {
...
+         min_level < ghext->num_levels();
+         min_level = max_level, max_level = min_level + 1) {
```
rhaas80: very confusing (even to me, who I think actually wrote this) since it really wants to do:

```
min_level = max_level;
max_level = max_level + 1;
```

and really we don't need to update `max_level` here (see comment below).

CarpetX/src/schedule.cxx
```diff
@@ -1746,21 +1778,20 @@ int Evolve(tFleshConfig *config) {
...
       // Find end of batch
-      int max_level = min_level + 1;
-
       while (max_level < ghext->num_levels()) {
```
rhaas80, line 1794: maybe easier code would be to have only:

```
for (int min_level = 0, max_level; min_level < ghext->num_levels(); min_level = max_level) {
```

in the loop header, then set:

```
max_level = min_level + 1
```

just before this loop? That way we keep all updates to `max_level` close to each other.

CarpetX/src/schedule.cxx
rhaas80: In only this code: no, I don't think it has to (since nothing in here ever does interpolation in time of any sort). However the Runge-Kutta based subcycling does make that assumption (namely that there's 2 fine steps to a coarse step). 

Not sure if I could envision a different factor (but that may just be b/c Carpet has this same restriction, so one has never been tempted).

CarpetX/src/schedule.cxx
```diff
@@ -1796,6 +1838,12 @@ int Evolve(tFleshConfig *config) {
 
       CycleTimelevels(cctkGH);
 
+      cctkGH->cctk_timefac = ghext->use_subcycling() ? (1 << min_level) : 1;
+      cctkGH->cctk_time =
+          ghext->use_subcycling()
+              ? cctkGH->cctk_delta_time * double(level_iteration)
+              : cctkGH->cctk_time + cctkGH->cctk_delta_time;
```
rhaas80, line 1845: the `cctkGH->cctk_time + cctkGH->cctk_delta_time` has been there before, however I wonder if the "new" code `cctkGH->cctk_delta_time * double(level_iteration)` is not actually the one to use in both cases since it avoids accumulating floating point errors (both `cctk_time` and `cctk_delta_time` being `CCTK_REAL` variables).

It would however forbid any use of adaptive time-stepping where `cctk_delta_time` changes continuously. Thoughts?

CarpetX/src/schedule.cxx
```diff
@@ -1805,11 +1853,26 @@ int Evolve(tFleshConfig *config) {
...
+      // current timestep
+      for (int level = min_level - 1; level >= 0; --level) {
+        rat64 coarse_iteration =
+            ghext->patchdata.at(0).leveldata.at(level).iteration;
+        if (coarse_iteration == level_iteration)
+          min_level = level;
+        else
+          break;
+      }
+      active_levels = make_optional<active_levels_t>(min_level, max_level);
```
rhaas80, line 1866: now here I think the comment that this better be all levels finer than `min_level` would allow for an easier assert? Namely this would mean that `max_level` is the maximum of all levels with any storage?

CarpetX/src/schedule.cxx
```diff
@@ -1805,11 +1853,26 @@ int Evolve(tFleshConfig *config) {
...
+
+      if ((!restrict_during_sync) &&
+          (ghext->patchdata.at(0).leveldata.at(max_level - 1).delta_iteration ==
```
rhaas80: ugly (b/c of the long series of `.` operations) and probably dicey in case patch `0` does not actually have `max_level` levels (that is if Lucas ever decides to make patch `0` to be not the Cartesian one or if we actually have mesh refinement in the curvilinear patches). What does that condition actually try to check? Bunch up the same levels we just stepped forward? it's not the set of levels we have decided have now caught up with each other. Couldn't `active_levels` contain levels for which this is not true? I am confused by this logic.

TestSubcycling/schedule.ccl
rhaas80: We should probably rename this test, since it does not check the full subcycling functionality (my bad, sorry).

TestSubcycleStepping/test/subcycle.par
rhaas80: Since the canary

```
    CCTK_REAL canary = 100 + 10 * cctk_iteration + 1 * cctk_level;
    iteration(pt.I) = canary;
```

is actually constant and we do not do anything that varies is space, two suggestions:

* can we get away with a single grid point?
* output only one grid point

WaveToyX/test/radiative-3levels.par
```diff
@@ -0,0 +1,55 @@
...
+
+WaveToyX::initial_condition = "Gaussian"
+WaveToyX::gaussian_width = 0.5
+WaveToyX::boundary_condition = "radiative"
+
+ODESolvers::method = "RK3"
+
+IO::out_dir = $parfile
+IO::out_every = 10
+IO::parfile_write = no
```
rhaas80, line 43: is missing the `IO::out_info` or whatever it's called setting to only output column labels (though those settings are currently ignored by CarpetX's output code).

WaveToyX/test/radiative-3levels.par
```diff
@@ -0,0 +1,55 @@
...
+
+CarpetX::out_metadata = no
+CarpetX::out_norm_vars = "all"
+CarpetX::out_norm_omit_unstable = yes
+CarpetX::out_norm_omit_sumloc_for_backward_compatibility = yes
```
rhaas80, line 48: since this is a new test, do we need to be backwards compatible?

rhaas80 (changes requested):

TestSubcycling/schedule.ccl
lwJi: I'll rename it to TestSubcycleStepping

TestSubcycleStepping/test/subcycle.par
lwJi: no. we need mesh refinement to test the subcycle stepping. we can't do mesh refinement with only one grid point. the fact the it's constant making it easier for us to tell if the subcycle stepping is correct or not

CarpetX/src/driver.hxx
lwJi: Make sense. Done

CarpetX/src/schedule.cxx
lwJi: Make sense. Done

CarpetX/src/schedule.cxx
lwJi: both points taken. PR updated

CarpetX/src/schedule.cxx
lwJi: Done

CarpetX/src/schedule.cxx
lwJi: It's a wrapper calling the regular `SyncGroupsByDirI`, but only apply to evolution variables

CarpetX/src/schedule.cxx
lwJi: fixed

CarpetX/src/schedule.cxx
lwJi: It sync only evolved GFs. I'll rename it to `SyncEvolvedGFs`

CarpetX/src/schedule.cxx
lwJi: fixed

CarpetX/src/schedule.cxx
lwJi: you are right. We don't. Fixed

CarpetX/src/schedule.cxx
lwJi: `cctk_levfac` are the same between subcyeling and non-subcycling. Where does `cctk_timefac` used?

CarpetX/src/schedule.cxx
lwJi: should I do anything about it? or it's fine for now

CarpetX/src/schedule.cxx
lwJi: We are trying to do restriction (and POSTRESTRICT) only when the finest level finish the current time stepping.
For example, 

```
----- ----- -----
             4/4
       4/4  -----
             3/4
 4/4  ----- -----
             2/4
       2/4  -----
             1/4
----- ----- -----
```

we don't want to call Restrict (and PostRestrict) when `level_iteration=4/4` at level 1, but at level 2 (the finest level). The reset of active_levels one line above make it contains all the level (all of them have `level_iteration=4/4`).

We are using `delta_iteration` to decide if we are stepping the finest level or not.

WaveToyX/test/radiative-3levels.par
lwJi: this test was generate before we modify CarpetX. Just to guarantee that we don't break anything when modify the schedule.cxx and so on.
All the other WaveTayX has this parameter and I think it's irrelevant to this pr

WaveToyX/test/radiative-3levels.par
lwJi: all the other tests in WaveToyX don't have that.

CarpetX/src/schedule.cxx
lwJi: I agree. but the original non-subycling implementation use `time+delta_time`. and if I changed here, it will break some unit test. It worth another separate PR to do the change

CarpetX/src/schedule.cxx
rhaas80: `cctk_timefac` is used in `CCTK_DELTA_TIME`:

```
src/include/cctk_core.h:#define CCTK_DELTA_TIME (cctk_delta_time/cctk_timefac)
```

we do actually update it:

```
CarpetX/src/schedule.cxx:  cctkGH->cctk_timefac = 1; // no subcycling
CarpetX/src/schedule.cxx:  cctkGH->cctk_timefac = sourceGH->cctk_timefac;
CarpetX/src/schedule.cxx:      cctkGH->cctk_timefac = ghext->use_subcycling ? (1 << min_level) : 1;
```

Are you using `CCTK_DELTA_TIME` in the `ODESolvers`? If so that is already a test for the value being set and being set correctly.

rhaas80: Liwei says out-of-band that `CCTK_DELTA_TIME` is used.

CarpetX/src/schedule.cxx
rhaas80: I think it is fine. Erik's comment would then apply to the `min_level` variable. I will add a sanity check.

CarpetX/src/schedule.cxx
lwJi: Done

CarpetX/src/schedule.cxx
lwJi: I think you are right. `cctkGH->cctk_delta_time * double(level_iteration)` works for both cases.
However, it would break some unit test because of an output issue. We should address this unification in a separate PR for clearness.

And yes, it won't work with continuous Adaptive time-stepping, but I don't think we will use it in near future.

CarpetX/src/schedule.cxx
lwJi: you are right. Now we only restrict when `max_level == ghext->numlevels()`

PR #373, opened 2026-01-05
CarpetX: Added option to exclude boundaries in 1D tsv output

lucass-carneiro:
This PR adds the option to exclude boundaries when saving 1D TSV files

### DIFF
--- a/CarpetX/param.ccl
+++ b/CarpetX/param.ccl
@@ -680,6 +680,9 @@ INT out_tsv_every "Output in TSV format every that many iterations" STEERABLE=al
   1:* :: "every that many iterations"
 } -1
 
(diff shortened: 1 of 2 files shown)


rhaas80 (approved):

eschnett:
Does this output just the outer boundary points, or does it output ghost points as well? If so, the documentation should be changed.

lucass-carneiro:
> Does this output just the outer boundary points, or does it output ghost points as well? If so, the documentation should be changed.

As far as my understanding goes, TSV files don't output ghosts, only outer boundaries (points starting with negative indices and extending beyond the cell size). This parameter excludes the outer boundaries in TSV output data.

PR #374, opened 2026-01-17
CarpetX: Check timers before changing their state

lucass-carneiro:
In some situations, adding `CarpetX` timer objects to scheduled functions may cause the flesh to issue warnings about the timer state, i.e., warnings against starting timers that are already started or stopping timers that are already stopped.

This PR checks the state of the underlying `Cactus` timer before starting or stopping them to prevent such warnings.

### DIFF
--- a/CarpetX/src/timer.cxx
+++ b/CarpetX/src/timer.cxx
@@ -16,7 +16,9 @@ Timer::Timer(const std::string &name)
 void Timer::start() {
 #pragma omp critical
   {
-    CCTK_TimerStartI(handle);
+    if (!CCTK_TimerIsRunningI(handle)) {
+      CCTK_TimerStartI(handle);
+    }
 #ifdef __CUDACC__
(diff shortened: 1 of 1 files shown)


lucass-carneiro:
Ticket is [here](https://bitbucket.org/einsteintoolkit/tickets/issues/2911/carpetx-check-timers-before-changing-their)

rhaas80 (changes requested):
This will not fix the issue since the problem is that in CarpetX a scheduled function is already being called by CarpetX in a multi-threaded manner so one would need one CCTK timer per thread to not have them not interfere.

lucass-carneiro:
This PR was created to fix an issue of timers being activated / decativated multiple times in multi-threaded contexts. This was an user error (on my part) of the `CarpetX` timers API.

No `Cactus` scheduled function that is run under `CarpetX`'s automatic parallelization (such as RHSs) can have `static` timers.

PR #377, opened 2026-03-12
Rhaas/sc

rhaas80:
@eschnett @lwJi @yosef-zlochower here's some sanity checks that will prevent checkpointing and interpolation when not possible.

### DIFF
--- a/CarpetX/src/interpolate.cxx
+++ b/CarpetX/src/interpolate.cxx
@@ -664,6 +664,11 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
     givis.at(v) = {gi, vi};
   }
 
+  rat64 min_level_iteration_used =
+      std::numeric_limits<decltype(min_level_iteration_used.num)>::max();
+  rat64 max_level_iteration_used =
+      std::numeric_limits<decltype(max_level_iteration_used.num)>::min();
+  bool requires_interpolation_in_time = false;
   for (const auto &patchdata : ghext->patchdata) {
     const int patch = patchdata.patch;
     for (const auto &leveldata : patchdata.leveldata) {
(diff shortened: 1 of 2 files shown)


rhaas80:
let me redo this...

rhaas80:
better

CarpetX/src/interpolate.cxx
```diff
@@ -676,6 +681,11 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
         const GridDesc grid(leveldata, mfp);
         // const int component = mfp.index();
 
+        min_level_iteration_used =
+            std::min(min_level_iteration_used, leveldata.iteration);
```
yosef-zlochower: This test seems to fail because min_level_iteration_used.numerator is long int max, and the test x.num * y.den overflows when den > 1
https://github.com/EinsteinToolkit/CarpetX/blob/e518a3cfedb1b2827add6ec3a1e2bbab8f775356/Arith/src/rational.hxx#L196

rhaas80:
Addressed reviewer comments.

CarpetX/src/interpolate.cxx
rhaas80: worked around by using `-1` to indicate the "unitilalized" case plus a couple of extra workarounds for MPI's `MPI_MAX` reduction. 

A "cleaner" solution would be to make the `<` comparing not overflow by doing repeating divisions with remainders until a decision has been reached (is slow) or using double-length numbers as intermediaries (that would be fast I think). Both are overkill for this single application though.

yosef-zlochower: I ran a test that confirmed that the code correctly errors out when checkpointing on unaligned time levels and that interpolation errors out if time interpolation is required. Checkpointing on aligned timelevels works and interpolations not requiring time interpolation also work

rhaas80: Great. Thanks for checking. That leaves as the only place that is a bit iffy the cast to `double` in the interpolator check, which reduces things to 53 bits of information (size of the mantissa of a `double`). However the iterations also need to fit into Cactus' `cctk_iteration` which is a `int` so 32bit only anyway and since refinement in time is restricted to powers of two we can exactly represent the `rat64` as a `double` for all cases that Cactus can actually encounter.

PR #378, opened 2026-03-13
Incresed maxncomps to accomdate more grid functions

annamalai-ps:


rhaas80 (changes requested):
typo in commit message, commit message should start with "CarpetX: "

80 is plenty. I assume the 640 bytes this allocates are not a problem on GPUs?

eschnett:
80 is a lot. How many do you actually need?

annamalai-ps:
Sorry I missed this. 80 is a lot, changed it to 30. I currently use 20 grid functions. I believe 30 should be plenty enough

PR #379, opened 2026-04-28
Interpolation caching with epochs

lucass-carneiro:
This PR is based on #369 and supersedes it by also including an epoch mechanism, that allows users to determine if their private caches require invalidation


rhaas80:
so I forget about #369 and review only this one or need to review both (one can update an existing pull request as well).

lucass-carneiro:
Yes 379 should supersede 369.

I did it in a separate branch because I was waiting for review on 369 and
didn’t want to interfere with that

CarpetX/src/driver.cxx
```diff
@@ -44,6 +44,8 @@ int ghext_handle = -1;
 amrex::AMReX *restrict pamrex = nullptr;
 std::unique_ptr<GHExt> ghext;
 
+std::atomic<CCTK_INT> carpetx_epoch{1};
```
rhaas80: I'd start from `0`, this is C++ code after all.

CarpetX/src/driver.hxx
```diff
@@ -483,6 +484,12 @@ struct GHExt {
 
 extern std::unique_ptr<GHExt> ghext;
 
+// Monotonically increasing counter. Incremented whenever the AMR grid
+// hierarchy is invalidated (regridding, recovery). Starts at 1.
```
rhaas80: the starting value does not have to be known to the users

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+
+// Create a cache of data required for interpolation
+struct InterpolationSetup {
+  using Container = amrex::AmrParticleContainer<3, 2>;
+  using Particle = Container::ParticleType;
+  using PinnedParticleTile = amrex::ParticleContainer_impl<
+      Particle, 0, 0, amrex::PinnedArenaAllocator>::ParticleTileType;
+  using ParticleTile = Container::ParticleTileType;
+
+  const CCTK_INT npoints{0};
```
rhaas80: I'd make this a `int` to match say https://einsteintoolkit.org/referencemanual/ReferenceManual.html#x1-115000 and reserve the `CCTK_FOO` types for "science" data and not just item counters.

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+  std::vector<char> symmetry_reflected_z;
```
rhaas80: why only ? Why not a `bool`?

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+  std::vector<std::shared_ptr<Container> > containers{};
```
rhaas80: how many containers? What's the index eg see CarpetX having stuff like:

```
std::vector<int> foo; // [rl]
```

to indicate that the vector will be indexed by the refinement level.

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+
+  InterpolationSetup(const CCTK_POINTER_TO_CONST cctkGH_,
+                     const CCTK_INT npoints,
+                     const CCTK_REAL *restrict const globalsx,
+                     const CCTK_REAL *restrict const globalsy,
+                     const CCTK_REAL *restrict const globalsz);
+};
+
+void InterpolateFromSetup(
```
rhaas80: "UsingSetup"? maybe? "Setup" is not the source of the interpolation

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+    const InterpolationSetup &setup, const CCTK_INT nvars,
+    const CCTK_INT *restrict const varinds,
+    const CCTK_INT *restrict const operations,
+    const std::vector<Arith::vect<Arith::vect<bool, 3>, 2> >
+        &outer_boundary_per_patch,
```
rhaas80: what do the indices mean? Is this the same ordering of indices (`[face][direction]`) as used in other places that deal with the same distiction? Eg Cactus' `cctk_bbox`? https://einsteintoolkit.org/usersguide/UsersGuide.html#x1-98000doc

CarpetX/src/interp.hxx
```diff
@@ -23,6 +32,34 @@ extern "C" CCTK_INT CarpetX_DriverInterpolate(
...
+    const CCTK_POINTER resultptrs_);
```
rhaas80: why have an underscore only here? This is a pointer or a pointer to point?

Since this is all C++ code (not aliased functions) could this be made more type save by using more expressive types?

CarpetX/src/interp.hxx
```diff
@@ -3,6 +3,15 @@
...
 #include <cctk.h>
 
+#include <defs.hxx>
+
+#include <AMReX_AmrParticles.H>
+#include <AMReX_Particles.H>
+
+#include <vector>
+#include <array>
+#include <memory>
```
rhaas80: this file needs to include `vect.hh` (or whatever) from `Arith` since you use `Arith::vect` below

CarpetX/src/interpolate.cxx
```diff
@@ -314,6 +305,21 @@ template <typename T, int order, int centering> struct interpolator {
 
       // Avoid points on boundaries
       const bool is_allowed = all(i >= i0_allowed && i < i1_allowed);
+
+      if (!is_allowed) {
+        CCTK_VWARN(0,
```
rhaas80: use `CCTK_ERROR` which is annotated to never return

CarpetX/src/interpolate.cxx
```diff
@@ -314,6 +305,21 @@ template <typename T, int order, int centering> struct interpolator {
...
+                   "Interpolation anchor is not allowed:\n"
```
rhaas80: please no multi-line error message. Those are hard to look for in output using regular expressions

rhaas80: mention why its not allowed (outside of box?)

CarpetX/src/interpolate.cxx
```diff
@@ -458,40 +460,49 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
   if (have_MultiPatch_GlobalToLocal2) {
     MultiPatch_GlobalToLocal2(npoints, globalsx, globalsy, globalsz,
                               patches.data(), localsx.data(), localsy.data(),
                               localsz.data());
   } else {
-    for (int n = 0; n < npoints; ++n)
+    for (int n = 0; n < npoints; ++n) {
       patches.at(n) = 0;
-    for (int n = 0; n < npoints; ++n)
+    }
```
rhaas80, line 470: has this file been run through `clang-format-NN` using the appropriate version of `clang-format-NN` and the correct style file?

CarpetX/src/interpolate.cxx
```diff
@@ -504,7 +515,8 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
   std::vector<CCTK_REAL> posx(npoints);
   std::vector<CCTK_REAL> posy(npoints);
   std::vector<CCTK_REAL> posz(npoints);
-#pragma omp simd
+
+#pragma omp parallel for simd
   for (int n = 0; n < npoints; ++n) {
     const int patch = patches.at(n);
     const amrex::Geometry &geom = ghext->patchdata.at(patch).amrcore->Geom(0);
```
rhaas80, line 521: are you sure that this AMReX call is thread save?

CarpetX/src/interpolate.cxx
```diff
@@ -552,14 +559,13 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
-  using ParticleTile = Container::ParticleTileType;
-  std::vector<Container> containers(ghext->num_patches());
+  containers.resize(ghext->num_patches());
   for (int patch = 0; patch < ghext->num_patches(); ++patch) {
     const PinnedParticleTile &pinned_particle_tile =
         pinned_particle_tiles.at(patch);
 
     const auto &restrict patchdata = ghext->patchdata.at(patch);
-    containers.at(patch) = Container(patchdata.amrcore.get());
+    containers.at(patch) = std::make_shared<Container>(patchdata.amrcore.get());
```
rhaas80: why shared pointers (and thus an extra indirection) if by-value worked before?

CarpetX/src/interpolate.cxx
```diff
@@ -643,6 +649,15 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
     }
 #endif
   }
+}
+
+void InterpolateFromSetup(
+    const InterpolationSetup &setup, const CCTK_INT nvars,
+    const CCTK_INT *restrict const varinds,
+    const CCTK_INT *restrict const operations,
+    const std::vector<vect<vect<bool, 3>, 2> > &outer_boundary_per_patch,
```
rhaas80: why call it `outer_boundary_per_patch` here when the code otherwise calls it `allowed_boundaries`?

CarpetX/src/interpolate.cxx
```diff
@@ -670,12 +685,21 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
-           ++pti) {
+      for (amrex::ParIter<3, 2> pti(*setup.containers.at(patch), level);
+           pti.isValid(); ++pti) {
         const MFPointer mfp(pti);
         const GridDesc grid(leveldata, mfp);
         // const int component = mfp.index();
 
+        // Compute per-box allowed_boundaries from per-patch outer boundary
+        // policy and AMReX bbox (which box faces touch the patch boundary)
+        vect<vect<bool, dim>, 2> ab;
```
rhaas80: what does `ab` stand for? A more descriptive variable name would be nice.

CarpetX/src/interpolate.cxx
```diff
@@ -670,12 +685,21 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
+        for (int f = 0; f < 2; ++f)
+          for (int d = 0; d < dim; ++d)
+            ab[f][d] = grid.bbox[f][d]
+                           ? outer_boundary_per_patch.at(patch)[f][d]
+                           : true;
```
rhaas80: so true for all inter-patch boundaries? Ie they can be used???

CarpetX/src/interpolate.cxx
```diff
@@ -713,36 +737,31 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
             switch (interpolation_order) {
             case 0: {
               const interpolator<CCTK_REAL, 0, 0b000> interp{
-                  grid,  gi,   vi,     patch,
-                  level, vars, derivs, bool(allow_boundaries)};
+                  grid, gi, vi, patch, level, vars, derivs, ab};
```
rhaas80: I note that the code before *did* use descriptive names. `allowed_boundaries`

CarpetX/src/interpolate.cxx
```diff
@@ -903,12 +918,39 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
+    for (int n = 0; n < setup.npoints; ++n) {
+      if (setup.symmetry_reflected_z[n]) {
         resultptrs[0][n] = -resultptrs[0][n];
         resultptrs[1][n] = +resultptrs[1][n];
       }
     }
   }
 }
+
+extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
```
rhaas80, line 943: this *is* providing an aliased function (why?) so yes, this must use Cactus types.

CarpetX/src/interpolate.cxx
```diff
@@ -903,12 +918,39 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
+                                    const CCTK_REAL *restrict const globalsx,
+                                    const CCTK_REAL *restrict const globalsy,
+                                    const CCTK_REAL *restrict const globalsz,
+                                    const CCTK_INT nvars,
+                                    const CCTK_INT *restrict const varinds,
+                                    const CCTK_INT *restrict const operations,
+                                    const CCTK_INT allow_boundaries,
+                                    const CCTK_POINTER resultptrs_) {
+  const InterpolationSetup setup(cctkGH_, npoints, globalsx, globalsy,
+                                 globalsz);
```
rhaas80: ok, the use here is why it takes Cactus types?

CarpetX/src/schedule.cxx
```diff
@@ -1111,6 +1120,7 @@ int Initialise(tFleshConfig *config) {
     RecoverGH(cctkGH);
     CCTK_Traverse(cctkGH, "CCTK_RECOVER_VARIABLES");
     CCTK_Traverse(cctkGH, "CCTK_POST_RECOVER_VARIABLES");
+    IncrementEpoch();
```
rhaas80: why here? And not in the part of the code that actually changes the grid?

CarpetX/src/schedule.cxx
```diff
@@ -1344,6 +1354,7 @@ int Initialise(tFleshConfig *config) {
           assert(!active_levels);
           active_levels = make_optional<active_levels_t>(
               first_modified_level, last_modified_level + 1);
+          IncrementEpoch();
```
rhaas80: I'd probably move taht closer to the top of this `if` block since nothing before is required to increment the epoch.

CarpetX/src/schedule.cxx
```diff
@@ -1740,6 +1751,7 @@ int Evolve(tFleshConfig *config) {
         assert(!active_levels);
         active_levels = make_optional<active_levels_t>(first_modified_level,
                                                        last_modified_level + 1);
+        IncrementEpoch();
```
rhaas80: same as before

CarpetX/src/driver.cxx
lucass-carneiro: The idea of starting from 1 is that user codes can store their epoch as 0, which would force a cache "recreation" at startup, which is what we want. Its more of a convenience.

CarpetX/src/driver.hxx
lucass-carneiro: Maybe, but:
1. This is just a comment, not an information being publicly broadcast somehow
2. As a user, you would like to start your cache counter with a value different than CarpetX's first so that you actually create a cache at startup.

We can change this, but suggestions are welcome

CarpetX/src/interp.hxx
lucass-carneiro: Done

CarpetX/src/interp.hxx
lucass-carneiro: So, `std::vector<bool>` is weird
1. is not memory contiguous
2. you can't take real references of it, it returns a proxy object.
3. Is not thread safe
4. and some other oddities:

https://isocpp.org/blog/2012/11/on-vectorbool
https://www.youtube.com/watch?v=OP9IDIeicZE

As a rule of thumb, I generally avoid it.

CarpetX/src/interp.hxx
lucass-carneiro: Done. Also removed some unused includes (as pointed out by my IDE's language server)

CarpetX/src/interpolate.cxx
lucass-carneiro: Done

CarpetX/src/interpolate.cxx
lucass-carneiro: No more multiline
Changed info to `Interpolation anchor is not allowed, as it lies outside of the interior region ...`

CarpetX/src/interpolate.cxx
lucass-carneiro: Yes. My IDE auto formats my code when I save it using clang-format and picking up the appropriate clang-format file.

CarpetX/src/interpolate.cxx
lucass-carneiro: *I'm* not sure, but, correct me if I'm wrong, this is part of Erik's code, just moved around, but still the same code he wrote.

In fact I think you can see the OG code [here](https://github.com/EinsteinToolkit/CarpetX/blame/bd6cc4c9be0497d6f44c82c471f0ed417c2b00b8/CarpetX/src/interpolate.cxx#L507)

CarpetX/src/interp.hxx
lucass-carneiro: This is function will be eventually called by `CarpetX_Interpolate` internally. I was just trying to make my life easier by keeping signatures and names consistent

CarpetX/src/interpolate.cxx
lucass-carneiro: This exported function was already provided as here. I only changed the guts to use an interpolation setup, but the interface is preserved. See [here](https://github.com/EinsteinToolkit/CarpetX/blob/bd6cc4c9be0497d6f44c82c471f0ed417c2b00b8/CarpetX/src/interpolate.cxx#L436)

CarpetX/src/interpolate.cxx
lucass-carneiro: This function is a piece of what `CarpetX_Interpolate` used to be. Because that took cactus types and I wanted to preserve the original code to the greatest extent possible, it also takes cactus types.

CarpetX/src/schedule.cxx
lucass-carneiro: I wanted to place all epoch increments on a single fie. The schedule seemed to be the logical place. Its easy to add and remove increments that way if necessary

CarpetX/src/interp.hxx
lucass-carneiro: One per patch. Added comment

CarpetX/src/interp.hxx
lucass-carneiro: In addition, the guts of this function come from pre-existing code. I wanted to keep variable names the same and change pre-existing code as little as possible.

CarpetX/src/interpolate.cxx
lucass-carneiro: For each patch, you configure if a boundary is allowed for interpolation (if it is outer) or not (it is interpatch). Thus `outer_boundary_per_patch` is true if it is an outer boundary, The name could be changed to smth like `patch_allowed_boundaries` but i think this is less understandable (a least to me). I'm willing to take suggestions

CarpetX/src/interpolate.cxx
lucass-carneiro: This is a little tricky. Originally, I have encountered runtime issues and accesses to stale data when passing particle `Container`s around using values. I believe that this works originally because you don't store the containers for any lifetime longer than the interpolation call itself (its all a single monolithic operation), so all is OK.

`Container` is explicitly non-copyable and AMReX documents that certain initialization steps must follow construction in a specific order. This + runtime issues when accessing cached particles told me the this type is intended to live at a stable address, not to be move-assigned into a pre-allocated slot. Promoting it to a shared ptr solved the issue and produced consistent results. We can, if you want, try and revert this to see if i can get the original bug to happen again

CarpetX/src/interpolate.cxx
lucass-carneiro: Ok, gave it some though and changed it to `allowed_boundary_policy`.
It's a policy that says weather or not a boundary can be used for interpolation.

CarpetX/src/driver.hxx
rhaas80: Ah, ok, that is a good idea. Well it's an `int` so `-1` would also be an option for user code (if starting from 0). But sure, this value of `1` then needs to be documented (if it is not already) so that code can actually make use of that detail.

CarpetX/src/interp.hxx
rhaas80:
```
repos/CarpetX$ git grep 'vector<bool>'
CarpetX/src/driver.hxx:  std::vector<bool> level_modified;
CarpetX/src/interpolate.cxx:  std::vector<bool> symmetry_reflected_z;
CarpetX/src/interpolate.cxx:  std::vector<bool> idxs(npoints, false);
CarpetX/src/io.cxx:std::vector<bool> find_groups(const char *const method,
CarpetX/src/io.cxx:  std::vector<bool> enabled(CCTK_NumGroups(), false);
CarpetX/src/io.cxx:        std::vector<bool> &enabled = *static_cast<std::vector<bool> *>(arg);
CarpetX/src/io.cxx:  const std::vector<bool> group_enabled = [&] {
CarpetX/src/io.cxx:    std::vector<bool> enabled(CCTK_NumGroups(), false);
CarpetX/src/io.cxx:    const std::vector<bool> input_group =
CarpetX/src/io.cxx:    const std::vector<bool> input_group =
CarpetX/src/io.cxx:  std::vector<bool> group_enabled(numgroups, false);
CarpetX/src/io.cxx:    std::vector<bool> &group_enabled =
CarpetX/src/io.cxx:        *static_cast<std::vector<bool> *>(callback);
CarpetX/src/io.cxx:      const std::vector<bool> group_enabled =
CarpetX/src/io.cxx:      const std::vector<bool> group_enabled =
CarpetX/src/io.cxx:      const std::vector<bool> group_enabled =
CarpetX/src/io.cxx:    const std::vector<bool> checkpoint_group = [&] {
CarpetX/src/io.cxx:      std::vector<bool> enabled(CCTK_NumGroups(), false);
CarpetX/src/io.cxx:    const std::vector<bool> checkpoint_group = [&] {
CarpetX/src/io.cxx:      std::vector<bool> enabled(CCTK_NumGroups(), false);
```

to late ...

CarpetX/src/interp.hxx
rhaas80: Right, I noticed that only later. I'd still pass at least `cctkGH` as a properly typed object. You have to cast it anyway, might just as well do it in the glue code.

CarpetX/src/interpolate.cxx
rhaas80: Which version? I am asking b/c the file `.clang-format` in `flesh` would indicate `clang-format-9` is to be used ("last tested with clang-format 9.0.1-16 by rhaas on 2021-01-23"). Running with "clang-format version 9.0.1 (https://github.com/muttleyxd/clang-format-static-binaries 5b56bb49b977b80cd18f44df4444db93e2d30f46)" on the modified files in this branch produces changes both in main and in the feature branch...

Looks like version 19 produces identical results (for those files but not all others) for me. Is that the one you used?

CarpetX/src/interpolate.cxx
rhaas80: You are not passing around `Container` objects though are you (they always stay in the vector)? 

`AmrParticleContainer` (which is `using AmrParticleContainer = AmrParticleContainer_impl<Particle<T_NStructReal, T_NStructInt>, T_NArrayReal, T_NArrayInt, Allocator, CellAssignor>`) is not copyable, but is moveable: see https://amrex-codes.github.io/amrex/doxygen/classamrex_1_1AmrParticleContainer__impl.html#pub-types So you should be able to move it. 

Are you sure that this was not (say) an issue that showed up if one did not (say) update the ParticleContainer after a regrid (ie after `amcore` has changed)?

rhaas80: Let's poke @eschnett and @lwJi about this, since they have both presumable worked with these `Containers`.

rhaas80: One could also use `containers.reserve(ghext->num_patches())` and then `std::vector::emplace` to construct the `Container` in place, in case one considers the move code "dodgy".

CarpetX/src/interpolate.cxx
rhaas80: still inconsistent with previous name used for the same concept. A choice has been made for the name before, unless there is very good reason to change it, I would not (since it forces everybody to re-learn the new name).

CarpetX/src/interpolate.cxx
rhaas80: True. But see comment about `cctkGH` earlier in this page.

CarpetX/src/interpolate.cxx
rhaas80: True. But see comment about `cctkGH` earlier in this page.

CarpetX/src/schedule.cxx
rhaas80: But is it also (not) error-prone? If this can be done automatically by the entity that does the regrid, then no one else needs to manually keep track.

rhaas80 (changes requested):
See inline comments.

CarpetX/src/interp.hxx
lucass-carneiro: So I can change this if you want me to, but i rather use a better type if we could. It may not matter that much now, but it may be that in the future we wish we did the right thing from the start.

CarpetX/src/interp.hxx
lucass-carneiro: Wait, what do you mean pass it as a properly typed object? Where? We are not passing `cctkGH` here

CarpetX/src/interpolate.cxx
lucass-carneiro: Humm here's what my `clang-format -- version` reports:

```
clang-format version 22.1.3
```

CarpetX/src/interpolate.cxx
lucass-carneiro: Ok, it's been a while since i did this and I don't remember all the details. While we wait for them, I will hack around and remove the shared ptr, temporarily, to see if I can reproduce the problem I was seeing. I will come back here once I have more

CarpetX/src/interp.hxx
lucass-carneiro: Ohhh wait, you mean `InterpolationSetup` should receive a `const cGH *restrict const` and not a `const CCTK_POINTER_TO_CONST cctkGH_` with cactus types since its a C++ struct, right?

CarpetX/src/interpolate.cxx
lucass-carneiro: Should I try and downgrade and run it again?

CarpetX/src/interp.hxx
rhaas80: well in that case it will at least be a uniform change to say `std::vector<bool_as_int>` without the awkward stage of having parts of the code use `std::vector<bool>` and parts something else.

CarpetX/src/schedule.cxx
lucass-carneiro: Hum, so looking at the code, maybe place the increment function immediately after the following places in `driver.cxx`

1. After line 1178
2. After line 1453
3. After line 1470

(basically every time `level_modified` is set to true or resized)

What do you think? Would that do it? (and then obv. remove it from the schedule)

CarpetX/src/interpolate.cxx
rhaas80: no, no need. There's already many versions in use. Your version's output agrees with version 19's output which seems to have been used in the past for these files.

CarpetX/src/interp.hxx
rhaas80: yes. I would *also* argue that the execution routine should again take such an pointer instead of relying on some internal copy of the same information (eg technically calling `CCTK_Abort` needs a `cctkGH`). So if you want to keep open the option of aborting the code, you will want to have that `cctkGH` handy.

CarpetX/src/interpolate.cxx
rhaas80: `#pragma omp simd` is not introducing parallel threads though. It just tells the compiler that the following loop is thread save. 

Presumably it is thread save though.

CarpetX/src/interpolate.cxx
rhaas80: (note that as usual, if you find another reviewer that is happy with that choice then it's a majority vote including the authors vote :-) ).

CarpetX/src/driver.cxx
lucass-carneiro: The cache now starts from 0

CarpetX/src/driver.hxx
lucass-carneiro: The cache now starts from 0

CarpetX/src/interp.hxx
lucass-carneiro: Name was changed

CarpetX/src/interp.hxx
lucass-carneiro: Docs were added with more clear info

CarpetX/src/interpolate.cxx
lucass-carneiro: Just leaving a comment, you were in fact right about it, I kinda added the `parallel`s too willy nilly. I restored the pragmas to what they used to be.

CarpetX/src/interp.hxx
lucass-carneiro: Hum so what you are saying is that both `InterpolationSetup` and `InterpolateUsingSetup` take a `const cGH *restrict const`?

Fair enough, but it would be unused (at least for now) in `InterpolateUsingSetup`. Also, I looked at `InterpolationSetup` again and the only use of the `gH` is an assertion to make sure we are in global mode... so its currently not very useful there too

rhaas80: True, though see what happened in the past: [4d9935b1](https://bitbucket.org/cactuscode/cactus/commits/4d9935b11d807dc47e6e69478c039e1dd507222c) "Cactus: fix problem with current_scheduled_function being a global." of [cactus](https://bitbucket.org/cactuscode/cactus) which had issues b/c `CCTK_CallFunction` was not (documented to) take a `cctkGH` (but all callers passed it thankfully).

Is it only unused b/c it internally uses a stored copy (or the global variable copy) of `ghext`?

CarpetX/src/interp.hxx
lucass-carneiro: Ok, I have changed this back to a vector of bools. I do feel icky about it but I'm not gonna die on this hill either :smile:, i rather just be practical about it

CarpetX/src/interpolate.cxx
lucass-carneiro: Ok, I restored `allowed_boundaries` :smile:

CarpetX/src/interp.hxx
lucass-carneiro: I can add the parameter, no problem, even if it is unused. is `[[maybe_unused]]` a thing we can rely on or do a `void(blah)` for now in the body?

And yes, internally only the global `ghext` is used. But again, this is one of the things that come about from the original, [here](https://github.com/EinsteinToolkit/CarpetX/blob/bd6cc4c9be0497d6f44c82c471f0ed417c2b00b8/CarpetX/src/interpolate.cxx#L448)

rhaas80: There's `CCTK_ATTRIBUTE_UNUSED` (not documented in the user's guide unfortunately, usually only useful for infrastructure code though). Use as say

```
const int cctki2_ialn CCTK_ATTRIBUTE_UNUSED = (cctki2_ialn_);
```

It is of course defined to be ` __attribute__((__unused__))` for gcc, but could be something else (or empty), if the compiler does not support that attribute.

CarpetX/src/interpolate.cxx
lucass-carneiro: Hum, Ok, it seems the "code no longer being broken when someone more senior looks at it" effect is taking hold.

I have reverted the code to not use a shared ptr and all carpetx tests pass. Not only that, the CapyrX tests that make use of the caching (with overlaps and mesh refinement) also pass and run as they should, so it was prob. what you said and the issue was something else that I fixed in the process of invalidating the cache at the proper times.

The downside of this is that we can no longer pass a `const InterpolationSetup &setup` to `InterpolateUsingSetup` because `amrex::ParIter<3, 2>` requires a non `const` particle container. Is this acceptable? If it is I can commit the change.

rhaas80: Maybe https://amrex-codes.github.io/amrex/doxygen/classamrex_1_1ParConstIter__impl.html can be used (Google search, no idea if this is useful or not, but it fits the `iterator` `const_iterator` pattern in C++)?

rhaas80: > Hum, Ok, it seems the "code no longer being broken when someone more senior looks at it" effect is taking hold.

https://github.com/EinsteinToolkit/CarpetX/pull/379#discussion_r3230349010 

"[Vorführeffekt](https://de-wikipedia-org.translate.goog/wiki/Vorf%C3%BChreffekt?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp)" though I guess technically "negated Vorführeffekt" 

And it's even in an English dictionary: https://en.wiktionary.org/wiki/Vorf%C3%BChreffekt

CarpetX/src/interp.hxx
```diff
@@ -42,15 +42,17 @@ struct InterpolationSetup {
 
   const int npoints{0};
   std::vector<bool> symmetry_reflected_z;
-  std::vector<std::shared_ptr<Container> > containers{}; // [patch]
+  std::vector<Container> containers{}; // [patch]
 
-  InterpolationSetup(const cGH *restrict const cctkGH, const CCTK_INT npoints,
+  InterpolationSetup([[maybe_unused]] const cGH *restrict const cctkGH,
```
rhaas80: yes i know that this is standard C++17, but: it's not used anywhere else in Cactus (or CarpetX for that matter). What is used is:

```
src/interpolate.cxx:  interpolator(const GridDescBase &grid, CCTK_ATTRIBUTE_UNUSED const int gi,
```

So it boils down to either confusing the "old timers" (those that predate C++17) or those that come in fresh from their C++ course which will have taught the actual standards conforming syntax. At least for C++ (for C one has to wait until C23 apparently).

CarpetX/src/interp.hxx
lucass-carneiro: I've added the `cctkGH` parameter. I decided to mark it with `[[maybe_unused]]`. This is C++ 17, so it should be fine for us to use it.

Also these are C++ interfaces, so we can do that

rhaas80 (commented):
down to nitpicking comments now.

CarpetX/src/interpolate.cxx
lucass-carneiro: > Maybe https://amrex-codes.github.io/amrex/doxygen/classamrex_1_1ParConstIter__impl.html can be used (Google search, no idea if this is useful or not, but it fits the `iterator` `const_iterator` pattern in C++)?

This was indeed useful.

I was able to restore `const` and eliminate the shared ptr. All carpetx and capyrx tests are still passing, so this seems to be OK

rhaas80 (approved):
Interesting. Once I submit a "request changes" or "approve" comment, GitHub won't let me punt the decision and put my state back to "neutral" via a comment only review.

So: personally I would use `CCTK_ATTRIBUTE_UNUSED` since (a) it's what is used in other places in CarpetX (and Cactus) and (b) its the same in C (and would be in Fortran if such a thing ever came up).

CarpetX/src/schedule.cxx
rhaas80: sure (since `level_modified` is not a class of its own). This is then similar to the `valid` tracking which is also done a set of parallel structures that are manipulated "close" to where the actual change happens

lucass-carneiro:
Thanks @rhaas80. This last commit should have all the modifications we discussed. When you can, take a look at it with a fine tooth comb to see if I missed anything.

lucass-carneiro:
@rhaas80, as asked by @eschnett this morning on the ETK call, `InterpolateUsingSetup` was promoted into a member function of `InterpolationSetup`. I took this opportunity and made `InterpolationSetup` into a `class`, with its internals `private`.

CarpetX/interface.ccl
```diff
@@ -121,6 +121,9 @@ void FUNCTION Interpolate(
   CCTK_POINTER IN resultptrs)
 PROVIDES FUNCTION Interpolate WITH CarpetX_Interpolate LANGUAGE C
 
+CCTK_INT FUNCTION GetEpoch()
+PROVIDES FUNCTION GetEpoch WITH CarpetX_GetEpoch LANGUAGE C
```
rhaas80, line 125: all need documentation

CarpetX/src/interpolate.cxx
```diff
@@ -643,6 +648,45 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
+ *
+ * When bbox[f][d] is true true, the box face is at the patch outer boundary. In
+ * this case we read allowed_boundaries for that face. If
+ * allowed_boundaries is true, that means some routine (i.e., BCs) has
+ * filled the ghost zones. When its false ghost zones are filled by inter-patch
+ * interpolation and we push the stencil anchor inward.
+ *
+ * When bbox[f][d] is false, the box face is interior to the patch, between
+ * AMReX boxes. In this case we return always true, because AMReX fill-patch
+ * operations guarantee these ghost zones are valid.
```
rhaas80: this is great. Can it go into `documentation.tex`?

CarpetX/src/interpolate.cxx
```diff
@@ -911,4 +968,33 @@ extern "C" void CarpetX_Interpolate(const CCTK_POINTER_TO_CONST cctkGH_,
...
+                                    const CCTK_INT nvars,
+                                    const CCTK_INT *restrict const varinds,
+                                    const CCTK_INT *restrict const operations,
+                                    const CCTK_INT allow_boundaries,
+                                    const CCTK_POINTER resultptrs_) {
+  const cGH *restrict const cctkGH = static_cast<const cGH *>(cctkGH_);
+
+  const InterpolationSetup setup(cctkGH, npoints, globalsx, globalsy, globalsz);
+
+  // Replicate OG behaviour:
```
rhaas80: "OG"?

rhaas80: "original" ?

PR #380, opened 2026-04-30
CarpetX: New parameter "out_tsv_combine_iterations"

eschnett:
Also rename "output_boundaries" to "out_tsv_ output_boundaries" because that is what that parameter does.

Needs test cases.

### DIFF
--- a/CarpetX/param.ccl
+++ b/CarpetX/param.ccl
@@ -717,6 +717,7 @@ USES STRING out_dir
 USES CCTK_INT out_every
 USES KEYWORD out_mode
 USES CCTK_INT out_proc_every
+USES KEYWORD out_fileinfo
 
 USES BOOLEAN checkpoint_ID
 USES INT checkpoint_every
--- a/CarpetX/param.ccl
+++ b/CarpetX/param.ccl
@@ -680,9 +680,21 @@ INT out_tsv_every "Output in TSV format every that many iterations" STEERABLE=al
   1:* :: "every that many iterations"
 } -1
 
-BOOLEAN output_boundary_points "Output outer boundary points in 1d TSV files" STEERABLE=always
+KEYWORD out_tsv_header "How to output TSV column header" STEERABLE=always
+{
+  "none" :: "no header"
+  "comment" :: "as comment (for backward compatibility)"
+  "plain" :: "plain TSV header"
+} "comment"
+
+BOOLEAN out_tsv_combine_iterations "Combine all iterations into a single TSV file" STEERABLE=recover
+{
+} no
+
+BOOLEAN out_tsv_output_boundary_points "Output outer boundary points in 1d TSV files" STEERABLE=always
 {
 } yes
+
 
 
 SHARES: Cactus
--- a/CarpetX/src/io_tsv.cxx
+++ b/CarpetX/src/io_tsv.cxx
@@ -293,7 +293,7 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
             file << sep << col++ << ":" << varname << ".real";
             file << sep << col++ << ":" << varname << ".imag";
           } else {
-            assert(0 && "Unexpected variable type");
+            CCTK_ERROR("internal error");
           }
         file << "\n";
       }
@@ -312,7 +312,7 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
             file << sep << varname << "_real";
             file << sep << varname << "_imag";
           } else {
-            assert(0 && "Unexpected variable type");
+            CCTK_ERROR("internal error");
           }
         file << "\n";
       }
@@ -397,7 +397,7 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
             file << sep << col++ << ":" << varname << ".real";
             file << sep << col++ << ":" << varname << ".imag";
           } else {
-            assert(0 && "Unexpected variable type");
+            CCTK_ERROR("internal error");
           }
         file << "\n";
       }
@@ -419,7 +419,7 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
             file << sep << varname << "_real";
             file << sep << varname << "_imag";
           } else {
-            assert(0 && "Unexpected variable type");
+            CCTK_ERROR("internal error");
           }
         file << "\n";
       }
@@ -782,7 +782,7 @@ void OutputTSV(const cGH *restrict cctkGH) {
                     out_fileinfo_val, out_header, output_boundary_points);
         break;
       default:
-        assert(0);
+        CCTK_ERROR("internal error");
       }
     }
   }
--- a/CarpetX/src/io_tsv.cxx
+++ b/CarpetX/src/io_tsv.cxx
@@ -22,8 +22,21 @@
 
 namespace CarpetX {
 
+enum out_header_t { header_none, header_comment, header_plain };
+static out_header_t get_out_header() {
+  DECLARE_CCTK_PARAMETERS;
+  if (CCTK_EQUALS(out_tsv_header, "none"))
+    return header_none;
+  if (CCTK_EQUALS(out_tsv_header, "comment"))
+    return header_comment;
+  if (CCTK_EQUALS(out_tsv_header, "plain"))
+    return header_plain;
+  CCTK_ERROR("Internal error");
+}
+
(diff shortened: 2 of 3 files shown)


rhaas80:
> Also rename "output_boundaries" to "out_tsv_ output_boundaries" because that is what that parameter does.

Hmm, @lucass-carneiro may need to comment, but if I recall correctly it's called "out_output_boundaries" because I complained in https://bitbucket.org/einsteintoolkit/tickets/issues/2910/carpetx-add-option-to-exclude-boundaries?iframe=true&spa=0#comment-69673234 about it having `tsv` in the name and that it would be nice to name it non-format specific in the hope that eventually all output formats would support such an option.

> Needs test cases.

ok

CarpetX/param.ccl
```diff
@@ -680,11 +680,23 @@ INT out_tsv_every "Output in TSV format every that many iterations" STEERABLE=al
   1:* :: "every that many iterations"
 } -1
 
-BOOLEAN output_boundary_points "Output outer boundary points in 1d TSV files" STEERABLE=always
+KEYWORD out_tsv_header "How to output TSV column header" STEERABLE=always
+{
+  "none" :: "no header"
+  "comment" :: "as comment (for backward compatibility)"
+  "plain" :: "plain TSV header"
+} "comment"
```
rhaas80, line 688: there's a `IOUtil` option `out_fileinfo` which is somewhat orthogonal to this in that it (should) control what goes into this header. 

"plain" is what the actual TSV standard wants (ie the first line of the TSV file is the column headers)?

CarpetX/src/io_tsv.cxx
```diff
@@ -145,25 +178,54 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
     varnames.push_back(CCTK_VarName(arraygroupdata.firstvarindex + vi));
 
   const std::string sep = "\t";
-  std::ofstream file(filename);
+  const std::ios::openmode mode =
+      combine_iterations ? std::ios::app : std::ios::out;
+  std::ofstream file(filename, mode);
```
rhaas80, line 262: this ignores `IOUtil::truncate_files`  (namely it never truncates an existing file if `combine_iterations` is true)

CarpetX/src/io_tsv.cxx
```diff
@@ -145,25 +178,54 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
...
-    if (cgroup.vartype == CCTK_VARIABLE_REAL ||
-        cgroup.vartype == CCTK_VARIABLE_INT)
-      file << sep << col++ << ":" << varname;
-    else if (cgroup.vartype == CCTK_VARIABLE_COMPLEX) {
-      file << sep << col++ << ":" << varname << ".real";
-      file << sep << col++ << ":" << varname << ".imag";
-    } else
-      assert(0 && "Unexpected variable type");
-  file << "\n";
+  if (file.tellp() == 0) {
```
rhaas80, line 268: when appending, this will leave the header inconsistent if say, the set of variables that are being output changes. 

Not sure if TSV allows headers in the middle of the file (probably not, since this would also means a change in number of columns).

CarpetX/src/io_tsv.cxx
```diff
@@ -145,25 +178,54 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
...
+      int col = 3;
+      for (const auto &varname : varnames)
+        if (cgroup.vartype == CCTK_VARIABLE_REAL ||
+            cgroup.vartype == CCTK_VARIABLE_INT) {
+          file << sep << col++ << ":" << varname;
+        } else if (cgroup.vartype == CCTK_VARIABLE_COMPLEX) {
+          file << sep << col++ << ":" << varname << ".real";
+          file << sep << col++ << ":" << varname << ".imag";
+        } else {
+          assert(0 && "Unexpected variable type");
```
rhaas80: probably should have made this a `CCTK_ERROR` when I added `complex` output, given that I point out that `assert()` can be optimized away :speak_no_evil:

CarpetX/src/io_tsv.cxx
```diff
@@ -145,25 +178,54 @@ void WriteTSVScalars(const cGH *restrict cctkGH, const std::string &filename,
...
+    }
+    case header_plain: {
+      file << "iteration" << sep << "time";
+      for (const auto &varname : varnames)
+        if (cgroup.vartype == CCTK_VARIABLE_REAL ||
+            cgroup.vartype == CCTK_VARIABLE_INT) {
+          file << sep << varname;
+        } else if (cgroup.vartype == CCTK_VARIABLE_COMPLEX) {
+          file << sep << varname << "_real";
+          file << sep << varname << "_imag";
```
rhaas80: why `_` here and `.` for "comment" style header? A limitation of the TSV file standard?

CarpetX/src/io_tsv.cxx
```diff
@@ -193,29 +257,61 @@ void WriteTSVArrays(const cGH *restrict cctkGH, const std::string &filename,
...
+        } else {
+          assert(0 && "Unexpected variable type");
+        }
+      file << "\n";
+      break;
+    }
+    }
+  }
```
rhaas80, line 429: same comments as for GF output

lucass-carneiro:
Just here to RT what Roland said.

The parameter was so named because  of his suggestion at that time. My original naming convention did include "tsv" in it.

rhaas80 (changes requested):

rhaas80:
any volunteers to try and give the `IO::truncate_files` issues a try? The corresponding code in (say) CarpetIOHDF5 looks like this:

```
  typedef std::map<string, vector<vector<vector<int> > > > filelist;
  static filelist created_files;

  filelist::iterator thisfile = created_files.find(basefilename);
  is_new_file = thisfile == created_files.end();
  truncate_file = is_new_file and IO_TruncateOutputFiles(cctkGH);
```

in the `DIdOutput` function which also handles the issue of multiple output for the same variable (and the same iteration and the same reflevel and mglevel).
so basically it will truncate the file on first output during this particular segment if requested to do so by `IO_TruncateOutputFiles(cctkGH)`.

rhaas80:
Ticket is here: https://bitbucket.org/einsteintoolkit/tickets/issues/2936

CarpetX/param.ccl
eschnett: Done.

`out_fileinfo` says which information to output, the new `out_tsv_header` says how to output it (as comment or plain). I added support for `out_fileinfo`.

CarpetX/src/io_tsv.cxx
eschnett: Fixed.

CarpetX/src/io_tsv.cxx
eschnett: Fixed.

CarpetX/src/io_tsv.cxx
eschnett: Things that appear in the header are usually translated into variable names by TSV readers. Variables with `.` in their name are weird.

I left the comments the way they were before.

CarpetX/src/io_tsv.cxx
```diff
@@ -22,26 +25,124 @@
...
+// Collect the information we may want to put into headers
+struct fileinfo_t {
+  std::string current_date, current_time;
+  std::string hostname;
+  std::string parameter_filename;
+};
+static fileinfo_t get_fileinfo() {
+  fileinfo_t fileinfo;
+
+  char current_date[1000];
```
rhaas80, line 60: ok, this is likely big enough for a date string in any imaginable locale.

rhaas80 (approved):

lucass-carneiro:
Ok so will `output_boundary_points` change name to `out_tsv_ output_boundaries` because the latter is more descriptive even though when I originally named it `out_tsv_boundaries` I had to change it [here](https://github.com/EinsteinToolkit/CarpetX/pull/373/commits/9c43c36ca5c40264dd871b9692c3102460f8c31a) due to backward naming consistency?

Just checking, because we are using `output_boundary_points` in Cottonmouth and will have to change it to the new name if this is the case. This will also interfere with the release, because cottonmouth test files will have to be changed and so on.

stevenrbrandt:
This name change would require a modification to all cottonmouth test par files.

rhaas80:
Ticket is: https://bitbucket.org/einsteintoolkit/tickets/issues/2936/carpetx-tsv-output-does-not-append-files

PR #382, opened 2026-05-18
Arith: Use checked integer arithmetic for rational numbers

eschnett:


Arith/src/rational.hxx
```diff
@@ -72,23 +74,28 @@ template <typename I> struct rational {
...
-    return rational(-x.num, x.den, no_normalize());
+    return rational(checked_neg(x.num), x.den, no_normalize());
   }
+  constexpr rational inv() const { return rational(den, num, no_normalize()); }
 
   friend constexpr rational operator+(const rational &x, const rational &y) {
-    return rational(x.num * y.den + x.den * y.num, x.den * y.den);
+    return rational(
+        checked_add(checked_mul(x.num, y.den), checked_mul(x.den, y.num)),
+        checked_mul(x.den, y.den));
```
rhaas80, line 87: these do not make things more readable, unfortuantely

rhaas80 (approved):
looks ok to me. 

less readable now unfortunately due to the function calls for math.

Arith/src/rational.hxx
eschnett: The alternative is to introduce a new type `checked_int` which is quite a bit more overhead.

PR #385, opened 2026-05-22
HydroBaseX: HydroBase feature porting, Aphi GF, conditional storage and inits.

Jimbo-Coder:
Move HydroBaseX closer to a true HydroBase/Carpet port by adding the missing optional initialization parameters: `initial_Y_e`, `*_Bvec`, `*_Avec`, `*_Aphi`, `*_temperature`, and `*_entropy`. These parameters declare which thorn provides each optional field, with HydroBaseX handling the built-in zero/one initialization cases.

Add Aphi as a HydroBaseX GF with AsterX-consistent vvv centering. This is the electric potential, which rounds out the EM support with Bvec, Avec, and Aphi all being treated consistently.

Add zeroing routines for Bvec, Avec, Aphi, temperature, and entropy, and add the Y_e one routine.

Rho, press, eps, and vel are always stored, while the remaining optional fields are now only stored conditionally, consistent with HydroBase's treatment.


Jimbo-Coder:
This PR to HydroBaseX would ideally go with #386 which makes similar changes to ADMBaseX, porting the conventions set forth by HydroBase/ADMBase respectively.

PR #387, opened 2026-05-25
Feature/any all

stevenrbrandt:
This PR allows the code despite a compiler bug in cuda 12.9.1 https://bitbucket.org/einsteintoolkit/tickets/issues/2922/carpetx-does-not-compile-with-cuda-129

It moves any/all to namespaces and fully qualifies their use in a few cases.

### DIFF
--- a/Arith/src/vect.hxx
+++ b/Arith/src/vect.hxx
@@ -810,6 +810,7 @@ template <typename T, int D> struct nan<vect<T, D> > {
   }
 };
 
+
 // Defined at namespace scope rather than as hidden friends inside vect
 // to work around an nvcc bug where ADL fails to resolve hidden-friend
 // all/any (observed with nvcc 12.6 through 13.1 + g++ 13).
@@ -830,7 +831,6 @@ any(const vect<T, D> &x) {
     r = r || x[d];
   return r;
 }
-
 } // namespace Arith
 namespace std {
 template <typename T, int D> struct equal_to<Arith::vect<T, D> > {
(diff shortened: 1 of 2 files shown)


rhaas80:
LGTM

eschnett:
I believe my recent PR that removes `using namespace std` fixes this. The problem is that C++23 defines a class `std::any` (and `std::all`), and this class is accidentally in scope because some header files use `using namespace std`. This is also the reason why things are failing only for `any` and `all`, and not for any of the other functions.

rhaas80:
I see. So the nvcc compiler on the failing cluster defaulted (or faultily used) to C++23 (since we should be asking for `-std=c++17`?

eschnett:
I think the _compiler_ is using C++17, but the _standard library_ is using C++23, probably simply because you're using a newer version of GCC (or nvcc).

rhaas80:
> I believe my recent PR that removes `using namespace std` fixes this. The problem is that C++23 defines a class `std::any` (and `std::all`), and this class is accidentally in scope because some header files use `using namespace std`. This is also the reason why things are failing only for `any` and `all`, and not for any of the other functions.

https://github.com/EinsteinToolkit/CarpetX/pull/381  is the PR? That pull request unfortunately  has build failures (missing `std::` in `setprecision` call in some io thorn. I will try and update it.

rhaas80 (commented):
Superseded by #381

rhaas80:
Superseded by https://github.com/EinsteinToolkit/CarpetX/pull/381

PR #388, opened 2026-06-26
Support vector grid functions in GF3D2

eschnett:


rhaas80 (approved):
LGTM. So the functionality in Arith that supports this:

```
Arith::masko_loadu(mask, &(*this)(I, n), other)
```

already existed?

eschnett:
No, `Arith` is not involved. This calls `&(*this)(I, n)`, which calls the constructor of `GF3D2` with `I, n` (which I added), takes the address, and passes it to `loadu`. All new functionality is in `loop.hxx`.

PR #390, opened 2026-07-16
Adapt to AMReX 26.02 Array4 API changes

lwJi:
AMReX 26.02 refactored `Array4` into a specialization of the new `ArrayND` class (AMReX-Codes/amrex#4900). This changed the public member API that CarpetX relies on:

- the `Dim3 begin`/`end` members became `IntVectND`s, so `vars.begin.x` etc. no longer compile;
- the public `jstride`/`kstride`/`nstride` members were replaced by a `get_stride<d>()` accessor;
- the new headers use C++20 `requires` clauses, so building against AMReX ≥ 26.02 needs `-std=c++20`.

## Changes

- **CarpetX**: use the `amrex::lbound`/`amrex::ubound` free functions instead of the `begin`/`end` members in `interpolate.cxx`, `io_tsv.cxx`, and `reduction.cxx`. These functions exist in all AMReX versions, so this compiles against both old and new AMReX. Note that `ubound` is inclusive while the old `end` member was exclusive; loop bounds and asserts are adjusted accordingly.
- **ODESolvers**: use `Array4::get_stride()` when building against AMReX ≥ 26.02, guarded by `AMREX_RELEASE_NUMBER` (the same pattern as the existing guard in `CarpetX/src/fillpatch.cxx`); the old stride members are kept for older AMReX.
- **scripts**: bump the CI configs from C++17 to C++20, required for AMReX ≥ 26.02 and harmless for older versions.

No behavior change is intended; this is a pure API adaptation.

## Tests

Built and ran the CarpetX testsuite against AMReX 26.07 (g++, macOS), which exercises the new (`AMREX_RELEASE_NUMBER >= 260200`) code path. The pre-26.02 code path in `ODESolvers` is textually unchanged and continues to compile against the AMReX shipped in the current CI containers.

These changes were originally developed and CI-tested (CPU/CUDA/ROCm/oneAPI, real32/real64) in lwJi/CarpetX#103.

### DIFF
--- a/scripts/actions-cpu-real32.cfg
+++ b/scripts/actions-cpu-real32.cfg
@@ -18,7 +18,7 @@ F90 = gfortran
 # Note: GCC 13.2 with -ftrivial-auto-var-init=pattern mis-compiles (?) C++ lambdas
 
 CFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -ftrivial-auto-var-init=pattern -g -march=x86-64-v3 -pie -pipe -std=gnu11
-CXXFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -g -march=x86-64-v3 -pie -pipe -std=gnu++17
+CXXFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -g -march=x86-64-v3 -pie -pipe -std=gnu++20
 FPPFLAGS = -traditional
 F90FLAGS = -fPIE -fcf-protection=full -fcray-pointer -ffixed-line-length-none -finit-character=65 -finit-integer=42424242 -finit-real=nan -fstack-clash-protection -fstack-protector-strong -g -march=x86-64-v3 -pie -pipe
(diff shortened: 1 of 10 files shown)


scripts/actions-cpu-real32.cfg
```diff
@@ -18,7 +18,7 @@ F90 = gfortran
 # Note: GCC 13.2 with -ftrivial-auto-var-init=pattern mis-compiles (?) C++ lambdas
 
 CFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -ftrivial-auto-var-init=pattern -g -march=x86-64-v3 -pie -pipe -std=gnu11
-CXXFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -g -march=x86-64-v3 -pie -pipe -std=gnu++17
+CXXFLAGS = -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS -fPIE -fcf-protection=full -fstack-clash-protection -fstack-protector-strong -g -march=x86-64-v3 -pie -pipe -std=gnu++20
```
rhaas80: C++20 is intentional?

eschnett (approved):

scripts/actions-cpu-real32.cfg
rhaas80: apparently so, since Liwei says "the new headers use C++20 requires clauses, so building against AMReX ≥ 26.02 needs -std=c++20." so we need to update all simfactory option lists to C++20

PR #403, opened 2026-09-11
Add support for Multi-step Runge-Kutta methods in `ODESolvers`

lucass-carneiro:
This PR adds support for The multi-step Runge-Kutta methods described [here](https://arxiv.org/abs/2603.05763) in `ODESolvers`

It also adds a new test thorn `TestMSRK` to test them and exemplifies their usage.

These methods require a history of previous RHSs to be stored in memory. In order to do that, we use the existing `Cactus` time level and storage infrastructure. Users declare that the RHS group may have (up to) a certain number of time levels in `interface.ccl` and explicitly require the storage in `schedule.ccl`.

Requiring extra storage regardless of `ODSolvers` method used, however, would be bad: Users using, say RK4, would always have to pay the extra memory cost with no benefit. In order to lazily allocate the required extra storage, `CarpetX` was modified to allow group storage increases and decreases. Memory is then lazily allocated if a multi-step method is required and the proper declarations are made in their thorn's CCL files.

Furthermore, some bugs in `CarpetX`'s time level cycling had to be fixed (currently no thorn is using this or testing time levels, so its not surprising we had some uncovered problems)


eschnett:
Why is it necessary to store the intermediate values as time levels? It would be easier to just allocate temporary variables, the same way ODESolver is already doing it for its temporaries.

Also, if you are using time levels, then we should be using the official flesh mechanisms to increase or decrease a group's storage, we should not introduce a new CarpetX-internal mechanism for this.

lucass-carneiro:
Hey Erik,

Thanks for the comments I was waiting for the CI jobs to pass before tagging you and Roland for review.

> Why is it necessary to store the intermediate values as time levels? It would be easier to just allocate temporary variables, the same way ODESolver is already doing it for its temporaries.

My idea here was to try and use an "official" cactus feature, that is, timelevels, to store the history for the RHS. I think, logically, that history fits well with the concept of time levels, it is data from a past iteration.

Of course, if you think having `ODESolvers` directly allocate RHS history when needed is a cleaner design I can implement that instead.

My only concern is to be able to allocate that storage lazily (only when a multistep method is selected) and have it not relocated every substep. Also, I may be wrong here, but when using the `ODESolvers` temporaries, rotating the past RHS levels would imply linear combination kernels (that's how I had these implemented before using timelevels) instead of just pointer swaps, which makes things a little less efficient than they could be.

> Also, if you are using time levels, then we should be using the official flesh mechanisms to increase or decrease a group's storage, we should not introduce a new CarpetX-internal mechanism for this.

Yes, I agree. In a (possibly failed) attempt to try and make this change not too big, I chose to introduce a `CarpetX`-only API on the hopes that, some day, when we want to, that can be wrapped into the flash API. If you'd like me to go ahead and do this now, I can do it.

Regardless, I think the new `TestTimeLevels` thorn is a good addition, even if we end up not using time levels for the multistep RK methods, since it creates a test case for a feature we technically have but was otherwise untested.

eschnett:
Look e.g. how the `RKF78` method is implemented in ODESolvers. It allocates a vector of temporaries, and the number of temporaries is flexible. If you were to drop in a different Butcher tableau, then the rest of the implementation would still work. I think this is exactly what you want, and there is no need to add any additional infrastructure.

lucass-carneiro:
> Look e.g. how the `RKF78` method is implemented in `ODESolvers`. It allocates a vector of temporaries, and the number of temporaries is flexible

Humm Ok, so [like this](https://github.com/EinsteinToolkit/CarpetX/blob/6c59144364fb5e65c7355d4a445530b8503460db/ODESolvers/src/solve.cxx#L1002)?

I can see the RHS history being stored as a `std::vector<statecomp_t>` in `ODESolvers`, but then I think I have a couple follow up questions:

1. This vector needs to be a global, or a static variable in the `ODESolvers_Solve` function, because we need to remember the value of some of the `ks` for previous iterations (that is the whole point of the methods). Are there any gotchas introducing a new global for that?
2. When initializing, regriding and recovering, we need to bootstrap these multistep methods with a regular RK method, as there is not enough past data available, or in the case of regridding, the values in the history become stale. The solution I had for this before this cleanup was a global grid scalar, that would be set to true in case bootstrapping was needed. Now we have a `epoch` mechanism, which I guess ODESolvers could query for some of that. Is this the design you have in mind?

PR #404, opened 2026-09-14
CarpetX: Make the checkpoint triggers independent of each other

eschnett:
Four mechanisms that can trigger a checkpoint: `checkpoint_ID`, `checkpoint_every`, `checkpoint_every_walltime_hours`, and `checkpoint_on_terminate`. Currently they share a single `last_checkpoint_runtime` that every checkpoint write updates. This can accidentally delay creating a checkpoint when `checkpoint_every_walltime_hours` is combined with other mechanisms, which can push it past a job's run time limit, leading to loss of progress.

This PR makes each checkpoint trigger mechanism independent, preventing the above.

Also, currently `checkpoint_every_walltime_hours` measures the time in between checkpoint writes, and this can accumulate a delay over along run time, especially if setting up initial conditions takes a long time.

This PR makes the wall time checkpoint trigger behave more closely to what one would expect from "every N hours". Checkpoints are now triggered at N hours, after the current iteration finishes.


CarpetX/doc/documentation.tex
```diff
@@ -323,6 +323,10 @@ \subsubsection{\texttt{checkpoint}}
...
 
+There are four mechanisms that can trigger a checkpoint, and they are mutually independent: \texttt{IO::checkpoint\_ID} (checkpoint the initial data), \texttt{IO::checkpoint\_every} (every so many iterations), \texttt{IO::checkpoint\_every\_walltime\_hours} (after so much wall time has passed), and \texttt{IO::checkpoint\_on\_terminate} (checkpoint before terminating). Each mechanism examines only its own state, and in particular a checkpoint produced by one of them never postpones another. Independently of how many mechanisms fire, \CarpetX\space writes at most one checkpoint per iteration -- a second checkpoint of the same iteration would hold the same data.
```
rhaas80: `\CarpetX\space` -> `\CarpetX{}`

CarpetX/doc/documentation.tex
```diff
@@ -323,6 +323,10 @@ \subsubsection{\texttt{checkpoint}}
...
+
+The wall time criterion \texttt{IO::checkpoint\_every\_walltime\_hours} is anchored to the start of the run rather than to the previous checkpoint. For an interval of $N$ hours it divides the run into slots $[0,N), [N,2N), [2N,3N), \ldots$ of wall time since the process started, and fires at the first opportunity that falls into a later slot than its previous firing did. The long-term rate is therefore exactly one checkpoint per $N$ hours, with no drift accumulating from the delay between a slot boundary and the next opportunity to checkpoint. A single time step longer than $N$ spans several slots but still produces one checkpoint rather than one per slot, since the skipped slots all denote the same state. The criterion is evaluated both during the evolution and once the initial data have been set up, so that a lengthy setup is not lost; since the wall time is measured per process, it restarts from zero after a recovery.
```
rhaas80, line 328: don't we have a `\code{FOO}` macro?

CarpetX/src/io.cxx
```diff
@@ -691,7 +693,67 @@ void Checkpoint(const cGH *const restrict cctkGH) {
...
+// checkpoint: slot boundaries lie at run times N, 2N, 3N, ... for an interval
+// of N seconds, and the trigger fires at the first evaluation point that falls
+// into a later slot than the previous firing did. The long-term rate is thus
+// exactly one checkpoint per N, with no drift accumulating from the delay
+// between a boundary and the next evaluation point.
+//
+// A single time step spanning several slots fires once rather than once per
+// slot, and the state then snaps to the current slot: the skipped slots all
+// denote the same simulation state, so the additional checkpoints would hold
+// identical data.
```
rhaas80: This comment (may) be redundant. Namely if (not a given, if we call once per level and once per iteration) we only call this function once per iteration, then certainly it cannot trigger more than once per iteration.

rhaas80 (approved):

CarpetX/doc/documentation.tex
eschnett: I'm using `\texttt` everywhere else already... I can change all the places, but maybe that should be a separate PR.

CarpetX/src/io.cxx
eschnett: If you checkpoint every 10 minutes, and a single iteration takes 30 minutes, then CarpetX will not try to catch up by writing 3 checkpoints in quick succession. Instead, it will write one, and then "forget" that it missed a few checkpoints ("snap to the current slot").

eschnett: That is, it wouldn't write 3 checkpoints during the next 3 iterations. It certainly wouldn't write 3 checkpoints for the same iteration (as you say), but it also doesn't re-trigger immediately at the next iteration.

